PR #29, opened 2017-08-10
Webcomponent setup

megheaiulian:
Setup eslint, travis-ci, web component tester, polymer


cosmoz-omnitable-group-row.html
```diff
@@ -63,7 +63,7 @@
 		},
 
 		_configureElement: function (element, column) {
```
nomego: Can't we just drop this? @plequang ?

plequang: Yes we can

PR #31, opened 2017-08-28
Issue 28

JaySunSyn:
Only a demo fix. Issue 28 has been closed already.

### DIFF
--- a/demo/full-demo.html
+++ b/demo/full-demo.html
@@ -100,14 +100,12 @@
 					</cosmoz-omnitable-column>
 					<cosmoz-omnitable-column title="Custom template"
 						value-path="name" sort-on="name" width="50px" flex="2"
-						data-template$="[[ customNameTemplate ]]">
+						data-template="[[ customNameTemplate ]]">
 					</cosmoz-omnitable-column>
 
-					<div slot="actions">
-						<template  is="dom-repeat" items="[[ _getRowItems(selectedItems.length) ]]">
-							<paper-button>[[ item ]]</paper-button>
-						</template>
-					</div>
+					<template  is="dom-repeat" items="[[ _getRowItems(selectedItems.length) ]]">
+						<paper-button slot="actions">[[ item ]]</paper-button>
+					</template>
 
 					<paper-button slot="actions" on-run="updateSelectedRows">
 						<iron-icon icon="editor:mode-edit"></iron-icon>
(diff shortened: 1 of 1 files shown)


JaySunSyn:
Can be merged now.

demo/full-demo.html
```diff
@@ -100,12 +100,14 @@
 					</cosmoz-omnitable-column>
 					<cosmoz-omnitable-column title="Custom template"
 						value-path="name" sort-on="name" width="50px" flex="2"
-						data-template="[[ customNameTemplate ]]">
+						data-template$="[[ customNameTemplate ]]">
```
nomego: Oh this highlights a potential pitfall.. @plequang 's naming of the property versus the element dataSet.. this should not be converted to an attribute anyway..

demo/full-demo.html
```diff
@@ -100,12 +100,14 @@
...
 					</cosmoz-omnitable-column>
 
-					<template is="dom-repeat" items="[[ _getRowItems(selectedItems.length) ]]">
-						<paper-button>[[ item ]]</paper-button>
-					</template>
+					<div slot="actions">
+						<template  is="dom-repeat" items="[[ _getRowItems(selectedItems.length) ]]">
+							<paper-button>[[ item ]]</paper-button>
+						</template>
+					</div>
```
plequang: Wrapping several buttons in a single slotted element will not work with `cosmoz-bottom-bar`, which will consider the whole div as a single element in the bottom bar.
For example, bottom-bar displays normally only up to 3 buttons/elements in the toolbar, then the rest in a dropdown menu. When wrapping several buttons in `<div slot="actions">`, the bottom bar will display more then 3 buttons (try it by selecting all rows in the demo).
You should set `slot="actions"` for each buttons.

JaySunSyn: Oh, great, I see!

demo/full-demo.html
JaySunSyn: omg, you are right, not sure why I did that...! Should I change the naming in the behavior or ignore the warning for now?

JaySunSyn: I see, there is already a new issue for renaming.

ref https://github.com/Neovici/cosmoz-omnitable/issues/32

plequang:
Demo looks good.
However, let's hope no-one will use something similar on large data sets.
(In demo, click on the "select all" checkbox, and watch the bottom-bar animation :-) )

PR #34, opened 2017-08-29
Issue 32 - renaming of data-template

JaySunSyn:


JaySunSyn:
Should we also rename something like this?
https://github.com/Neovici/cosmoz-omnitable/blob/master/cosmoz-omnitable-column-behavior.html#L179

plequang:
Maybe it would be cleaner to rename all usage of data-template and dataTemplate ?
That would make the code more coherent.

For example, you can set a specific data/cell template 
(https://github.com/Neovici/cosmoz-omnitable/blob/master/docs/documentation.md#custom-data-templates-in-light-dom):
```html
<cosmoz-omnitable-column>
   <template id="data-template">
...
```
This could be renamed to `cell-template`
Same apply to `edit-data-template`.

**However**, I just realized using `id` as an attribute to identify the templates is a bad idea.
`id` is meant to be unique within a document

So I think it would be better to change the way we define and locate the templates by using:
```html
<template class="header">
<template class="cell">
<template class="edit-cell">
```

What do you think @JaySunSyn ?

JaySunSyn:
I think thats a good idea :) Working on it..

JaySunSyn:
please check @plequang

PR #36, opened 2017-09-01
Issue 22 - creation of time column

JaySunSyn:
In Progress.

After I get a OK from you guys on current changes, I'll continue with cleanup, documentation, linter fixes and minor refactors.

@nomego please check.

### DIFF
--- a/cosmoz-omnitable.html
+++ b/cosmoz-omnitable.html
@@ -231,7 +231,7 @@
 							<paper-listbox class="dropdown-content menu" selected="{{ groupOn }}" attr-for-selected="data-group-on">
 								<paper-item data-group-on=""><i style="white-space: nowrap">[[ _('No grouping', t) ]]</i></paper-item>
 								<template is="dom-repeat" items="[[ columns ]]" as="column">
-									<paper-item data-group-on$="[[ column.groupOn ]]" hidden$="[[ !column.groupOn ]]">[[ column.title ]]</paper-item>
+									<paper-item data-group-on$="[[_orAttr(column, 'name', 'groupOn')]]" hidden$="[[ !column.groupOn ]]">[[ column.title ]]</paper-item>
 								</template>
 							</paper-listbox>
 						</paper-dropdown-menu>
--- a/cosmoz-omnitable.html
+++ b/cosmoz-omnitable.html
@@ -133,7 +133,7 @@
 						list,
 						sortFunc;
 
-					list = data.data;
+						list = data.data;
 					sortFunc = !data.reverse ? ascSortFunc(data.sortOn) : descSortFunc(data.sortOn);
 					list.sort(sortFunc);
 					postMessage({
--- a/cosmoz-omnitable.js
+++ b/cosmoz-omnitable.js
@@ -467,21 +467,23 @@
 
 			}, this);
 		},
-
-		_getColumn({ name, columnAttribute = 'name', alternativeColumnAtrribute } = {}) {
+		/*
+		 * Returns a column based on a name
+		 */
+		_getColumn({ attributeValue, attribute = 'name', alternativeAttribute } = {}) {
 			var col,
 				altCol;
 
-			if (!name) {
+			if (!attributeValue) {
 				return;
 			}
 
 			this.columns.find(column => {
-				if (column[columnAttribute] === name) {
+				if (column[attribute] === attributeValue) {
 					col = column;
 					return true;
 				}
-				if (alternativeColumnAtrribute && column[alternativeColumnAtrribute] === name) {
+				if (alternativeAttribute && column[alternativeAttribute] === attributeValue) {
 					altCol = column;
 				}
 			}, this);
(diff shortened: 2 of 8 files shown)


cosmoz-omnitable.js
```diff
@@ -1004,14 +1010,21 @@
...
 
-			if (sortOn && !sortOn.valuePath) {
+				if (!column) {
+					return;
+				}
 				this.sortOn = {
-					valuePath: sortOn,
-					descending: false
+					valuePath: column ? column.sortOn : null,
+					descending: sortOn.descending || false,
```
nomego: This was set to `false` to reset to ascending when changing sort column, which I think is expected from a UX perspective.

nomego: Should rather be column "default" if we move it into column-behavior.

cosmoz-omnitable.html
```diff
@@ -133,7 +133,7 @@
 						list,
 						sortFunc;
 
-					list = data.data;
+						list = data.data;
```
nomego: Revert

cosmoz-omnitable.html
```diff
@@ -231,7 +231,7 @@ <h3 class="groupRow-label">
 							<paper-listbox class="dropdown-content menu" selected="{{ groupOn }}" attr-for-selected="data-group-on">
 								<paper-item data-group-on=""><i style="white-space: nowrap">[[ _('No grouping', t) ]]</i></paper-item>
 								<template is="dom-repeat" items="[[ columns ]]" as="column">
-									<paper-item data-group-on$="[[ column.groupOn ]]" hidden$="[[ !column.groupOn ]]">[[ column.title ]]</paper-item>
+									<paper-item data-group-on$="[[_orAttr(column, 'name', 'groupOn')]]" hidden$="[[ !column.groupOn ]]">[[ column.title ]]</paper-item>
```
nomego: Just `column.name`

cosmoz-omnitable.js
```diff
@@ -467,41 +467,40 @@
...
+				altCol;
 
-			if (!this.groupOn) {
+			if (!attributeValue) {
 				return;
 			}
-			this.columns.some(function (column) {
-				if (column.groupOn === this.groupOn) {
+
+			this.columns.find(column => {
```
nomego: When replacing `some()` with `find()`, do `var col = this.columns.find(..)`
Also we can move the declaractions here - `var altCol, col = this.columns.find(..)`

JaySunSyn:
Cool, will do!

Regarding time column:

The date column is actually a date-time column if you set the userFormat property respectivley.

```
<cosmoz-omnitable-column-date name="date" user-format="MM/DD/YYYY HH:mm:ss" title="Timestamp" value-path="date" sort-on="date" group-on="date">
```

It is actually even a time column, if you set `user-format="HH:mm:ss"`. The only thing that would not work here is sorting & grouping by time-only values. It would still take the whole date-time values into consideration. This is where the time column would stick out..

Do I miss something?

cosmoz-omnitable.js
JaySunSyn: It is actually on false here by default.

From outsite of the element,

- Somebody wants to change the sorting:

`this.$.omnitable.sortOn = 'date'` // works, descending defaulted to false

- Somebody change the sort direction:

`this.$.omnitable.sortOn = {valuePath: 'date', descending: true}`

nomego: Well what happens if you sort on date, then again (to make it descending) and then sort on something else, doesn't it still sort descending on the new property?

nomego:
Well the date column can handle times, but the header won't allow the user to enter a time in the filter range.

JaySunSyn:
Should I rather try then to extend the date-column to support datetime values also in the filter or really crate a new date-time column?

nomego:
I think the cleanest is to create a datetime column and share common functionality in a behavior.
We will likely design the header different, may even use other elements, and it is nice with sane defaults for userFormat.

cosmoz-omnitable.js
JaySunSyn: Yes, so if you want to sort on a property with a descending sort order, you have to assign sortOn the respective Object. Otherwise, you just set the String 'sortOn = time'.

```
this.$.omnitable.sortOn = {valuePath: 'time', descending: true}
```

Alternatively, I could add a `descending` property on `omnitable`?

JaySunSyn:
Pefect, will do!

cosmoz-omnitable.js
nomego: Well actually, as in the other comments, I don't think we should overwrite sortOn with strings and objects in different scenarios.

But what happens if you run the 'full-demo', sort on Id (check that it's Ascending), sort on id again (check that it switches to Descending), and then sort on Boolean. What direction does it sort then?

JaySunSyn:
In favor of:

- https://github.com/Neovici/cosmoz-omnitable/pull/38
- https://github.com/Neovici/cosmoz-omnitable/pull/39
- https://github.com/Neovici/cosmoz-omnitable/pull/40

PR #38, opened 2017-09-05
groupOn/sortOn refactor with name-support

JaySunSyn:
@nomego please check

### DIFF
--- a/cosmoz-omnitable.html
+++ b/cosmoz-omnitable.html
@@ -147,13 +147,13 @@
 		<div class="mainContainer">
 			<div class="header" id="header">
 				<div class="selectAllCheckbox" hidden$="[[ !_showCheckboxes ]]">
-					<paper-checkbox on-change="onAllCheckboxChange" hidden$="[[ _noData ]]">
+					<paper-checkbox on-change="onAllCheckboxChange" hidden$="[[ !_dataIsValid ]]">
 					</paper-checkbox>
 				</div>
 				<cosmoz-omnitable-header-row columns="[[ visibleColumns ]]"></cosmoz-omnitable-header-row>
 			</div>
 			<div class="tableContent" id="tableContent">
-				<template is="dom-if" if="[[ _noData ]]">
+				<template is="dom-if" if="[[ !_dataIsValid ]]">
 					<div class="tableContent-empty">
 						<iron-icon icon="icons:announcement"></iron-icon>
 						<div>
@@ -162,7 +162,7 @@
 						</div>
 					</div>
 				</template>
-				<template is="dom-if" if="[[ _tooStrictFilterInfo ]]">
+				<template is="dom-if" if="[[ _filterIsTooStrict ]]">
 					<div class="tableContent-empty">
 						<iron-icon icon="icons:announcement"></iron-icon>
 						<div>
--- a/cosmoz-omnitable.html
+++ b/cosmoz-omnitable.html
@@ -242,7 +242,7 @@
 								<paper-item on-tap="_onSortColumnActivate"><i style="white-space: nowrap;">[[ _('No sorting', t) ]]</i></paper-item>
 								<template is="dom-repeat" id="sortColumns" items="[[ columns ]]" as="column">
 									<paper-item on-tap="_onSortColumnActivate" hidden$="[[ !column.sortOn ]]">
-										<span>[[ column.title ]]</span>&nbsp;<span>[[ _getSortDirection(column, sortOn.*) ]]</span>
+										<span>[[ column.title ]]</span>&nbsp;<span>[[ _getSortDirection(column, _sortOn.*) ]]</span>
 									</paper-item>
 								</template>
 							</paper-listbox>
--- a/cosmoz-omnitable.html
+++ b/cosmoz-omnitable.html
@@ -231,7 +231,7 @@
 							<paper-listbox class="dropdown-content menu" slot="dropdown-content" selected="{{ groupOn }}" attr-for-selected="data-group-on">
 								<paper-item data-group-on=""><i style="white-space: nowrap">[[ _('No grouping', t) ]]</i></paper-item>
 								<template is="dom-repeat" items="[[ columns ]]" as="column">
-									<paper-item data-group-on$="[[ column.groupOn ]]" hidden$="[[ !column.groupOn ]]">[[ column.title ]]</paper-item>
+									<paper-item data-group-on$="[[ column.name ]]" hidden$="[[ !column.groupOn ]]">[[ column.title ]]</paper-item>
 								</template>
 							</paper-listbox>
 						</paper-dropdown-menu>
--- a/cosmoz-omnitable.html
+++ b/cosmoz-omnitable.html
@@ -227,8 +227,8 @@
 			<div class="footer">
 				<div class="footer-controls">
 					<div class="footer-control">
-						<paper-dropdown-menu no-animations vertical-align="bottom" horizontal-align="left" id="groupOnSelector" label="[[ _('Group on', t) ]]">
-							<paper-listbox class="dropdown-content menu" selected="{{ groupOn }}" attr-for-selected="data-group-on">
+						<paper-dropdown-menu vertical-align="bottom" horizontal-align="left" id="groupOnSelector" label="[[ _('Group on', t) ]]">
+							<paper-listbox class="dropdown-content menu" slot="dropdown-content" selected="{{ groupOn }}" attr-for-selected="data-group-on">
 								<paper-item data-group-on=""><i style="white-space: nowrap">[[ _('No grouping', t) ]]</i></paper-item>
 								<template is="dom-repeat" items="[[ columns ]]" as="column">
 									<paper-item data-group-on$="[[ column.groupOn ]]" hidden$="[[ !column.groupOn ]]">[[ column.title ]]</paper-item>
@@ -237,8 +237,8 @@
 						</paper-dropdown-menu>
 					</div>
 					<div class="footer-control">
-						<paper-dropdown-menu no-animations vertical-align="bottom" horizontal-align="right" id="sortOnSelectorMenu" label="[[_('Sort on', t)]]">
-							<paper-listbox class="dropdown-content menu" id="sortOnSelector" selected="{{ _sortOnSelectorSelected }}">
+						<paper-dropdown-menu vertical-align="bottom" horizontal-align="right" id="sortOnSelectorMenu" label="[[_('Sort on', t)]]">
+							<paper-listbox class="dropdown-content menu" slot="dropdown-content" id="sortOnSelector" selected="{{ _sortOnSelectorSelected }}">
 								<paper-item on-tap="_sortItemTapped"><i style="white-space: nowrap;">[[ _('No sorting', t) ]]</i></paper-item>
 								<template is="dom-repeat" id="sortColumns" items="[[ columns ]]" as="column">
 									<paper-item on-tap="_sortItemTapped" hidden$="[[ !column.sortOn ]]" data-name$="[[ column.name ]]">
--- a/cosmoz-omnitable.html
+++ b/cosmoz-omnitable.html
@@ -238,11 +238,11 @@
 					</div>
 					<div class="footer-control">
 						<paper-dropdown-menu no-animations vertical-align="bottom" horizontal-align="right" id="sortOnSelectorMenu" label="[[_('Sort on', t)]]">
-							<paper-listbox class="dropdown-content menu" id="sortOnSelector" on-iron-items-changed="_updateSelectedSortIndex" selected="{{ _sortOnSelectorSelected }}">
-								<paper-item on-tap="_onSortColumnActivate"><i style="white-space: nowrap;">[[ _('No sorting', t) ]]</i></paper-item>
+							<paper-listbox class="dropdown-content menu" id="sortOnSelector" selected="{{ _sortOnSelectorSelected }}">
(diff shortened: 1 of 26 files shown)


cosmoz-omnitable.js
```diff
@@ -171,7 +188,7 @@
 		},
 
 		observers: [
-			'_sortOnChanged(sortOn.*)',
+			'_createSortOnObject(sortOn, descending)',
```
nomego: _sortOn should be computed

cosmoz-omnitable.html
```diff
@@ -231,18 +231,18 @@ <h3 class="groupRow-label">
...
+									<paper-item data-group-on$="[[ column.name ]]" hidden$="[[ !column.groupOn ]]">[[ column.title ]]</paper-item>
 								</template>
 							</paper-listbox>
 						</paper-dropdown-menu>
 					</div>
 					<div class="footer-control">
 						<paper-dropdown-menu vertical-align="bottom" horizontal-align="right" id="sortOnSelectorMenu" label="[[_('Sort on', t)]]">
-							<paper-listbox class="dropdown-content menu" slot="dropdown-content" id="sortOnSelector" on-iron-items-changed="_updateSelectedSortIndex" selected="{{ _sortOnSelectorSelected }}">
-								<paper-item on-tap="_onSortColumnActivate"><i style="white-space: nowrap;">[[ _('No sorting', t) ]]</i></paper-item>
+							<paper-listbox class="dropdown-content menu" slot="dropdown-content" id="sortOnSelector" selected="{{ _sortOnSelectorSelected }}">
```
nomego: How many of these event listeners can we get rid of if we use `attr-for-selected` instead?
Seems like the index isn't helping much.

cosmoz-omnitable.html
```diff
@@ -231,18 +231,18 @@ <h3 class="groupRow-label">
...
+								<paper-item on-tap="_sortItemTapped"><i style="white-space: nowrap;">[[ _('No sorting', t) ]]</i></paper-item>
 								<template is="dom-repeat" id="sortColumns" items="[[ columns ]]" as="column">
-									<paper-item on-tap="_onSortColumnActivate" hidden$="[[ !column.sortOn ]]">
-										<span>[[ column.title ]]</span>&nbsp;<span>[[ _getSortDirection(column, sortOn.*) ]]</span>
+									<paper-item on-tap="_sortItemTapped" hidden$="[[ !column.sortOn ]]" data-name$="[[ column.name ]]">
```
nomego: Is data-name ever used? Seems like you use event.model instead?

cosmoz-omnitable.js
```diff
@@ -64,20 +64,37 @@
...
 			},
 
+			descending: {
+				type: Boolean,
+				value: false
+			},
+
+			descendingText: {
+				type: String,
+				computed: '_computeDescendingString(_sortOn)',
```
nomego: This only needs to observe `_sortOn.descending`

cosmoz-omnitable.js
```diff
@@ -949,100 +985,22 @@
...
-						valuePath: column.sortOn,
-						descending: false
-					};
-				}
-
-				// Force dropdown menu to refresh `selectedItemLabel`
-				this._sortOnSelectorSelected = 0;
-				this._sortOnSelectorSelected = selected;
-			}
+		_sortItemTapped: function (event) {
```
nomego: This event-handler actually make _sortOn and sortOn get out of sync.
We should only need one place to set _sortOn (computed).

cosmoz-omnitable.js
```diff
@@ -215,6 +234,25 @@
 
 			this.columns.forEach(function (column) {
 				this.listen(column, 'filter-changed', '_onColumnFilterChanged');
+
+				// Check correct column setup
+				if (column.name && columnNames[column.name]) {
+					console.error('The name attribute needs to be unique among all columns! Not unique on column', column.name);
+				} else
+
+				if (!column.name){
```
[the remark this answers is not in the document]
JaySunSyn: Forgot to npm install..

cosmoz-omnitable.js
```diff
@@ -949,100 +985,22 @@
...
-				}
-
-				// Force dropdown menu to refresh `selectedItemLabel`
-				this._sortOnSelectorSelected = 0;
-				this._sortOnSelectorSelected = selected;
-			}
+		_sortItemTapped: function (event) {
+			var column = event.model ? event.model.column : undefined,
+				descending = this._sortOn && this._sortOn.columnName === column.name ? !this._sortOn.descending : false;
+			this.set('_sortOn', {columnName: column.name, descending: descending});
```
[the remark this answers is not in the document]
JaySunSyn: got rid of the whole _sortOn object.

cosmoz-omnitable.js
```diff
@@ -253,6 +291,28 @@
...
+
+		_showSortDirection(column, sortOnChange) {
+			var sortOn = sortOnChange.base;
+			if (!column || !sortOn) {
+				return false;
+			}
+			return sortOn.columnName === column.name;
+		},
+
+		_reselectDropdownItem() {
```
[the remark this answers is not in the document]
JaySunSyn: I moved _decendingText to the label of the dropdown. What do you think?

image

cosmoz-omnitable.js
JaySunSyn: got rid of _sortOn

JaySunSyn:
@nomego please check. 
After ok, I want to add some documentation and do some cleanup before merge.

cosmoz-omnitable.js
```diff
@@ -209,12 +223,33 @@
 				return;
 			}
 
-			// TODO: Un-listen from old columns ?
-
 			this.columns = columns;
 
+			// Check if column names are set and unique
+			this.columns.filter((column, i) => {
```
nomego: When there are multiple chained functions, use the following style:

```js
this.columns
  .filter(...) {
    ...
  }
  .forEach(...) {
     ....
  }
 ```

This makes the data flow easier to read vertically.

File: cosmoz-omnitable.js [hunk 10]
```diff
@@ -209,12 +223,33 @@
...
+				var name = column.name;
+				if (!name) {
+					console.error('The name attribute needs to be set on all columns! Missing on column', column.title, column);
```
nomego: Return here and skip the name check below

cosmoz-omnitable.js
```diff
@@ -209,12 +223,33 @@
...
+				}
+				return name && columnNames.indexOf(name) !== columnNames.lastIndexOf(name) && columnNames.indexOf(name) === i;
```
nomego: Why is comparison with `i` required?

cosmoz-omnitable.js
```diff
@@ -209,12 +223,33 @@
...
+			// TODO: Un-listen from old columns ?
 			this.columns.forEach(function (column) {
 				this.listen(column, 'filter-changed', '_onColumnFilterChanged');
+
+				if (!column.name){
+					// No name set; Try to set name attribute via valuePath
+					if (!valuePathNames) {
+						valuePathNames = this.columns.map(c => c.valuePath);
+					}
+					var hasUniqueValuePath = valuePathNames.filter(n => n === column.valuePath).length === 1;
```
nomego: How come we do different unique checks (indexOf + lastIndexOf vs filter) ?
Gut feeling says indexOf-approach is more performant?

cosmoz-omnitable.js
```diff
@@ -209,12 +223,33 @@
...
+					if (hasUniqueValuePath && !columnNames.indexOf(column.valuePath) === -1) {
```
nomego: Negation typo, please add a test for this

cosmoz-omnitable.js
```diff
@@ -64,20 +64,32 @@
...
+			 * - columnName: item's value path to sort on
 			 * - descending: a boolean indicating of sort is done in descending order.
 			 */
-			sortOn: {
-				type: Object,
-				value: null
-			},
 
 			// Index of the selected item in the sortOn listbox
 			_sortOnSelectorSelected: {
```
nomego: Drop in cleanup

cosmoz-omnitable.js
nomego: Actually, we might want to just take help from ES6 - https://developer.mozilla.org/sv-SE/docs/Web/JavaScript/Reference/Global_Objects/Set

JaySunSyn:
```
var a = ['name', 'date', 'name'];

func(a) // returns ["name"]

var func = function (arr) {
	return arr.filter((e, i) => {
            return arr.indexOf(e) !== arr.lastIndexOf(e) && arr.indexOf(e) == i
       })
}
```

without the comparison with i:

```
func(a) // returns ["name", "name"];
```

JaySunSyn: I thought about using Set but then I would have to get the intersection between `columnNames` and `columnNameSet` to find the duplicates which wouldn't make it much easier..

nomego: Well this filter logic assumes that any first column is the right one and anyone declared after that is wrong.
Since all we're doing with the columns is log an error that it's a duplicate.
I wouldn't mind if it logged out both columns causing the duplicate.

What do you think ?

JaySunSyn: Thats true...

cosmoz-omnitable.js
```diff
@@ -184,7 +170,9 @@
 
 		observers: [
 			'_dataChanged(data.*)',
-			'_debounceSortItems(sortOn, descending)'
+			'_debounceSortItems(sortOn, descending, filteredGroupedItems)',
+			'_debounceGroupItems(filteredItems, groupOn)',
```
nomego: This should probably observe groupOnColumn since it's needed in _groupItems.
Also _groupItems probably doesn't need (this.)groupOn at all but rather groupOnColumn.groupOn?

cosmoz-omnitable.js
```diff
@@ -64,46 +73,46 @@
...
+				value: ''
 			},
 
 			/**
-			 * Column that matches the current `groupOn` value.
+			 * The column that matches the current `groupOn` value.
 			 */
 			groupOnColumn: {
 				type: Object,
-				readOnly: true,
```
nomego: Why can't this be readOnly if it's computed?

cosmoz-omnitable.js
```diff
@@ -467,58 +499,26 @@
...
 		},
 
 		_updateVisibleColumns: function () {
 			var visibleColumns = this.columns.slice();
 
 			if (this.groupOn) {
 				visibleColumns = visibleColumns.filter(function (column) {
-					if (column.groupOn === this.groupOn) {
-						this._setGroupOnColumn(column);
+					if (column.name === this.groupOn) {
```
nomego: This could have actually just become a `return column.name !== this.groupOn;` one-line callback

cosmoz-omnitable.js
```diff
@@ -562,96 +559,85 @@
...
-				}
+			if (!groupOnColumn || !groupOnColumn.groupOn) {
+				this.filteredGroupedItems = this.filteredItems;
+				this._groupsCount = 0;
+				return;
+			}
 
-				this.filteredItems.forEach(function (item, index) {
-					var groupOnValue = groupOnColumn.getComparableValue(item, groupOn);
+			this.filteredItems.forEach(function (item) {
```
nomego: Fingers itch to make this a reduce() :)

cosmoz-omnitable.js
JaySunSyn: Its automatically readOnly if computed.

cosmoz-omnitable.js
```diff
@@ -290,109 +217,42 @@
...
+		_computeFilterIsTooStrict(dataIsValid, visibleItemsLength) {
+			return dataIsValid && visibleItemsLength < 1;
 		},
 
-		/**
-		 * Helper method to remove an item from `data`.
-		 * @param  {Object} item Item to remove
-		 * @return {Object} item removed
-		 */
-		removeItem: function (item) {
```
JaySunSyn: @nomego @plequang I didn't find a file here which used removeItem(). Is it used anywhere else?

cosmoz-omnitable.js
```diff
@@ -709,24 +611,11 @@
...
-		_computeIcon: function (item, expanded) {
-			return expanded ? 'expand-less' : 'expand-more';
-		},
-
-		_computeShowCheckboxes: function (dataIsValid, hasActions) {
-			return dataIsValid && hasActions;
-		},
-
-		// TODO(pasleq): re-implement expand/collapse of single item
-		toggleExtraColumns: function (event) {
```
JaySunSyn: Removed.

cosmoz-omnitable.js
```diff
@@ -709,24 +611,11 @@
 				}.bind(this));
 			}
 		},
-
-		// TODO(pasleq): re-implement expand/collapse of single item
-		_computeIcon: function (item, expanded) {
```
JaySunSyn: Removed.

cosmoz-omnitable.js
plequang: This is used externally.
Actually, this is just a convenience function. Users of `cosmoz-omnitable` could achieve same functionality by updating the array passed the `data` property.
This quite useful when implementing "Remove selected items" functionality:  after you've called your backend, you can easily update the omnitable so that it does not display these items anymore.

JaySunSyn: thx for explaining. re-added

Issue #41, opened 2017-09-13
Sort-on and group-on dropdowns do not update labels when columns use computed titles


plequang:
# Description

Using columns having a computed title, for example:

```html
<cosmoz-omnitable...
  <cosmoz-omnitable-column title="[[  computeTitle(someOtherProperty) ]]"...
```

When the title of this column changes because it is recomputed by the function, then the group on and sort on dropdown selector are not updated with the new title.

And when the table is grouped, the group header is also not updated.

## Steps to reproduce

In the full demo at http://127.0.0.1:8081/components/cosmoz-omnitable/demo/full-demo.html, the `Amount` and `Group` columns have translated title, using translatable behavior from https://github.com/Neovici/cosmoz-i18next

Changing the locale to "en" should change the "Group on" dropbox labels, but it doesn't.

Also, if you group by "Group" the table, then change the locale, the group header is not updated.

plequang:
dropdowns do not get updated because column's title changes are not propagated to `dom-repeat` elements.
Also, there is an issue in paper-dropdown-menu causing the current selected item's label to not be updated https://github.com/PolymerElements/paper-dropdown-menu/issues/197

PR #43, opened 2017-09-13
Alternate fix to #42 for issue #41

nomego:
There doesn't seem to be a clean way to handle this and I'm a bit puzzled how we can't get the subproperty changes to notify dom-repeat through the built-in binding/notification system, so I guess we have to live with a "renotify" solution for now.

But I do think using attributes and a MutationObserver is a bit overkill and not precise enough, although we would get a more generalized re-notify solution (which I'm not sure we need at this point).

Also I think having generic `translationsChanged` observers in combination with regular translation observers will at best cause overhead and in worst case generate race conditions or other weird behavior.

Finally I try to avoid async/debounce/timeout as workarounds - it kind of begs for async issues down the line.

I think I found a more general issue with the translations in neovici/cosmoz-i18next@c0c7681

Let me know what you think but I really think we should focus on merging #38 first.
(This PR is intended to showcase another path and actually has pseudo code for the sort on selectedItem label. :exclamation: )

### DIFF
--- a/cosmoz-omnitable-column-behavior.html
+++ b/cosmoz-omnitable-column-behavior.html
@@ -10,7 +10,8 @@
 		properties: {
 
 			title: {
-				type: String
+				type: String,
+				notify: true
 			},
 
 			valuePath: {
--- a/cosmoz-omnitable.js
+++ b/cosmoz-omnitable.js
@@ -211,11 +211,12 @@
 
 			// TODO: Un-listen from old columns ?
 
-			this.columns = columns;
-
-			this.columns.forEach(function (column) {
+			columns.forEach(function (column) {
 				this.listen(column, 'filter-changed', '_onColumnFilterChanged');
+				this.listen(column, 'title-changed', '_onColumnTitleChanged');
 			}, this);
+
+			this.set('columns', columns);
 
 			this._updateVisibleColumns();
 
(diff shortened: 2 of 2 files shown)


plequang:
> There doesn't seem to be a clean way to handle this and I'm a bit puzzled how we can't get the subproperty changes to notify dom-repeat through the built-in binding/notification system, so I guess we have to live with a "renotify" solution for now.

I've spent some time on this and couldn't find another solution. This does not mean there isn't.
Maybe with Polymer 2.x data binding improvements https://www.polymer-project.org/2.0/docs/about_20#data-system

> But I do think using attributes and a MutationObserver is a bit overkill and not precise enough, although we would get a more generalized re-notify solution (which I'm not sure we need at this point).

I tried your solution as well but then I thought you would prefer the MutationObserver :smiley:

> Also I think having generic translationsChanged observers in combination with regular translation observers will at best cause overhead and in worst case generate race conditions or other weird behavior.

Beware that the paper-dropdown-menu issue is not linked to title change issue.
If translations for "No grouping" changes, but not title change, then you will not refresh the paper-dropdown-menu. Another solution would be to have properties for all the translatable labels, but I think we decided to not do that long time ago.

> I think I found a more general issue with the translations in Neovici/cosmoz-i18next@c0c7681

This does not solve the timing issue. The column still get notified of translation changes after omnitable.
This because `created` is called from top to bottom.

The correct fix to that is to ensure translatable elements are registered from bottom to top in the `translationElements` array. Hadn't time to implement that though.

PR #44, opened 2017-09-15
Test Preparation & small improvments

JaySunSyn:
- some small improvements found while creating tests 
- verifyColumnSetup() created for testing override purposes

### DIFF
--- a/cosmoz-omnitable.js
+++ b/cosmoz-omnitable.js
@@ -266,8 +266,6 @@
 			this._debounceFilterItems();
 		},
 
-<<<<<<< HEAD
-=======
 		_onColumnTitleChanged: function (event) {
 			var column = event.target,
 				columnIndex;
@@ -288,7 +286,6 @@
 			}
 		},
 
->>>>>>> upstream/master
 		// Handle selection/deselection of a group
 		_onGroupCheckboxChange: function (event) {
 			var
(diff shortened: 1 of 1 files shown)


nomego:
Sorry for late review, looks good but please rebase and I'll have another look.

cosmoz-omnitable.js
```diff
@@ -255,18 +265,6 @@
 			this._debounceFilterItems();
 		},
 
-		_onColumnTitleChanged: function (event) {
```
plequang: This should not be removed, as this is part of fix for #47
Beware however while rebasing that a new version of this function has been committed (see ac4ce3fcd48dbbe361437a648e112b4b70a41c02)

cosmoz-omnitable.js
```diff
@@ -224,6 +225,10 @@
 
 		_scalingUp: false,
 
+		_computeVisibleColumns(columns, groupOn) {
+			return groupOn ? columns.filter(c => c.name !== this.groupOn) : columns;
```
plequang: You should not return the `columns` property reference here (when groupOn is false), but rather a shallow copy, to prevent unexpected modification of the `columns` array when modifying the `visibleColumns` array.

JaySunSyn: I think computed properties are readOnly but if we force somehow to change visibleColumns it's better I guess...

Issue #20, opened 2017-04-19
Support for hidden columns


plequang:
If a column has the hidden attribute set, then it should not be rendered in the table.

PR #46, opened 2017-09-15
Support for hidden columns [issue-20]

JaySunSyn:
In Progress. https://github.com/Neovici/cosmoz-omnitable/issues/20

@plequang  Please let me know if this strategy is fine for you.

I exclude columns from `omnitable.columns` which also excludes it from grouping/sorting. (Or would you rather only change the visibility?)

Also, do we need to observe changes in the hidden attribute? 

Thanks

### DIFF
--- a/cosmoz-omnitable-column-behavior.html
+++ b/cosmoz-omnitable-column-behavior.html
@@ -48,6 +48,16 @@
 			},
 
 			/**
+			 * Indicates whether the user has engaged with the
+			 * header/filter component of the column
+			 */
+			headerFocused: {
+				type: Boolean,
+				value: false,
+				notify: true
+			},
+
+			/**
 			 * Indicate that the column is loading/performing work
 			 */
 			loading: {
@@ -154,6 +164,17 @@
 				type: Boolean,
 				reflectToAttribute: true,
 				observer: '_hiddenChanged'
+			},
+			/**
+			 * Index of this column in the list of displayed columns (excluding disabled/hidden columns).
+			 */
+			columnIndex: {
+				type: Number
+			},
+
+			preferredDropdownHorizontalAlign: {
+				type: String,
+				computed: '_computePreferredDropdownHorizontalAlign(columnIndex)'
 			},
 
 			_dataTemplatizer: {
@@ -263,6 +284,10 @@
 
 		_getDefaultStyleModule: function () {
 			return;
+		},
+
+		_computePreferredDropdownHorizontalAlign: function (columnIndex) {
+			return columnIndex === 0 ? 'left' : 'right';
 		},
 
 		getString: function (item, valuePathArg) {
--- a/cosmoz-omnitable-column-behavior.html
+++ b/cosmoz-omnitable-column-behavior.html
@@ -150,6 +150,13 @@
 				observer: '_cellTemplateChanged'
 			},
 
+			hidden: {
+				type: Boolean,
+				value: false,
+				notify: true,
+				reflectToAttribute: true
+			},
+
 			_dataTemplatizer: {
 				type: Object,
 				value: null
--- a/cosmoz-omnitable-column-behavior.html
+++ b/cosmoz-omnitable-column-behavior.html
@@ -11,7 +11,7 @@
 
 			title: {
 				type: String,
-				notify: true
+				observer: '_titleChanged'
 			},
 
 			valuePath: {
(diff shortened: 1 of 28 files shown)


nomego:
Yes this pretty much circumvents Polymer's observers and properties.
I would suggest to add a `hidden` property to the column behavior and observe changes to that property instead. It could be reflected to an attribute to keep them in sync both ways.

cosmoz-omnitable.js
```diff
@@ -798,6 +798,15 @@
...
 			return false;
 		},
+        /**
+         * Helper method to get a node's hidden attribute as a boolean.
+         * @param {any} node The node.
+         * @returns {Boolean} True if node's hidden attribute is set `hidden` or `hidden="true"`.
+         */
+		_hiddenAttr(node) {
+			var v = node.getAttribute('hidden');
+			return (v ? v.toLowerCase() : v) === 'false' ? false : v === '' ? true : !!v;
```
plequang: This is wrong actually.
`hidden` is a boolean attribute, so the only valid values are either the empty string or the string `hidden`.
The correct code would be to use `node.hasAttribute`

JaySunSyn:
@nomego please check

cosmoz-omnitable.js
```diff
@@ -330,10 +331,14 @@
 			this._debounceFilterItems();
 		},
 
+		_debounceUpdateColumns() {
+			this.debounce('updateColumns', this._updateColumns);
```
plequang: debounce here seems legit, because when several columns have the hidden attribute, the `cosmoz-column-hidden-changed` event will be fired several times.
But then it raises question about the call to `_updateColumns`  in the node observer added in `created`.

plequang:
Additionally to my review commit, I just realized that if we calls multiple times `_updateColumns`, then we must ensure that we do not listen to `filter-changed` events on the same column several times.

cosmoz-omnitable.js
JaySunSyn: The created callback gets only called once, so no need to debounce that in my opinion..

JaySunSyn: Ok, now I get you. Yes, makes sense! So we only have one call to `_updateColumns` instead of two from 1) `created` and from 2) `debounceUpdateColumns`.

JaySunSyn:
Good catch @plequang! I now fire a custom `cosmoz-column-filter-changed` event instead of listening imperatively.

cosmoz-omnitable.js
```diff
@@ -210,6 +211,7 @@
 					return;
 				}
 
+				// this._debounceUpdateColumns();
 				this._updateColumns();
```
plequang: This need to be debounced if we want to collapse calls to _updateColumns into a single call during initialization.

cosmoz-omnitable-column-behavior.html
```diff
@@ -316,7 +326,15 @@
...
-		}
+		},
+
+		_hiddenChanged: function () {
+			this.fire('cosmoz-column-hidden-changed', { column: this });
+		},
+
+		__filterChanged: function () {
+			this.fire('cosmoz-column-filter-changed', { column: this });
+		},
```
plequang: OK so now we are entering some synchronizatrion/timing hell.
The `__filterChanged` observer is triggered when the column is readied.
But at that time, the node observer in omnitable `created` callback has not been called yet, so the `columns` property of omnitable is not defined.

This means under some circumstances (when data is avalaible immediately), we will try to filter items, but columns is undefined, raising an error.
(this is what happens in the tests).

JaySunSyn: Thanks. Please check my fix.
Another possibility would be to add/remove all `cosmoz-column-*` listeners while looping through `_updateColumns` as it has been before. What do you think?

plequang: Your fix looks correct.

The problem we had with listening for `*-changed` events on columns directly was that this event was not raised for `title` property when using polymer option `suppressBindingNotifications` (see https://github.com/Neovici/cosmoz-omnitable/issues/47).
This was causing translation issues, and that's why I changed to a custom event fired when the title was changed.
Same issue could occur with `hidden` and `filter` as these might be changed from above.

You're right, we could add/remove listeners of `cosmoz-column-*` events in `_updateColumns`.
Actually, I don't know what is best, between having 3 listeners per columns (and non bubbling events), or 3 listener for all columns (and bubbling events). If this is just a matter of code design, I don't care much. But are there performance implications ?
If there are no performance issues, seems like having 3 listeners for all columns is simpler to manage (no need to listen/unlisten when columns is hidden/unhidden - enabled/disabled).

So I'm fine with your fix, but if you feel we should do differently, it's up to you.

Issue #10, opened 2016-11-28
Row highlighting


nomego:
Omnitable should be able to highlight rows for different reasons.

1. Clicking on a row (without clicking on a link or checkbox) we might want to highlight it in a color(?)
2. Selecting an item with the checkbox should also color the whole row in (another?) color?
3. Programmatically highlighting one or more affected rows according to a related component's actions / selected items

plequang:
This issue should be clarified a bit I think.

Should highlighted rows be distinct from selected items ? Or should we have a property like `highlight-selected-items` to control that behavior.

Do we need single row highlighting ?

Also, currently the row templates are instantiated by cosmoz-grouped-list, which provide the following model to templates:
```
{
  selected: false, 
  item: {},
  folded: false, // for grow row
  expanded: false, // for item row
}
```

If we want to add `highlighted` to this model, I think we will need to update cosmoz-grouped-list, meaning highlighted rows should be handled in that component.

JaySunSyn:
Did single row highlighting with this PR https://github.com/Neovici/cosmoz-grouped-list/pull/9

I could add multiple row highlighting either here or in grouped-list.

My first thought was, when a highlighted row gets selected, it might should get "de-highlighted". Rather than the other way around. But not sure about this.. Its like level 1, 2, 3 while level 2 is not 1 & 2 but only 2. It's not exactly like this herre but I guess you know what I mean.

JaySunSyn:
Can be closed.

PR #49, opened 2017-09-19
Add support for highlighted items [issue-10]

JaySunSyn:
See issue https://github.com/Neovici/cosmoz-omnitable/issues/10

Please merge https://github.com/Neovici/cosmoz-grouped-list/pull/9 first!

### DIFF
--- a/cosmoz-omnitable.html
+++ b/cosmoz-omnitable.html
@@ -182,7 +182,6 @@
 											<paper-checkbox checked="{{ selected }}" on-change="_onItemCheckboxChange"></paper-checkbox>
 										</div>
 										<cosmoz-omnitable-item-row
-											on-tap="_itemRowTapped"
 											columns="[[ visibleColumns ]]"
 											selected="{{ selected }}"
 											expanded="{{ expanded }}"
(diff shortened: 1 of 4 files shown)


JaySunSyn:
@nomego In this PR I do not consider questions like:
- can a selected row get highlighted?
- should a highlighted row that gets selected still be highlighted?

nomego:
Great, but I think we should be able to configure the behavior if it's multi or single highlight with a property.

JaySunSyn:
@nomego what about checking if the input is an array or an object and deciding on multi/single based on that?

JaySunSyn:
@nomego please check

cosmoz-omnitable.html
```diff
@@ -173,15 +173,16 @@
...
 						<cosmoz-grouped-list-template id="itemTemplate">
 							<template>
 								<div class="item-row-wrapper">
-									<div class$="[[ _getItemRowClasses(selected) ]]">
+									<div selected$="[[selected]]" class="itemRow" highlighted$="[[highlighted]]">
 										<div class="selectItemCheckbox" hidden$="[[ !_showCheckboxes ]]">
 											<paper-checkbox checked="{{ selected }}" on-change="_onItemCheckboxChange"></paper-checkbox>
 										</div>
 										<cosmoz-omnitable-item-row
+											on-tap="_itemRowTapped"
```
plequang: You can remove that event handler for now.

PR #51, opened 2017-09-26
Name attribute in documentation

dotpointer:
Adding name attribute to documentation.

### DIFF
--- a/docs/documentation.md
+++ b/docs/documentation.md
@@ -37,6 +37,9 @@ provide type specific functionalities.
 </cosmoz-omnitable-column>
 ```
 
+The `name` attribute is mandatory and must be distinct among all columns.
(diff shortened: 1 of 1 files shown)


docs/documentation.md
```diff
@@ -31,6 +31,7 @@ provide type specific functionalities.
 
 ```html
 <cosmoz-omnitable-column
+      name="name"
```
plequang: Would be nice to add a small text below about the name attribute, indicating it is mandatory and must be distinct among all columns

dotpointer: I've edited the documentation. Please have a look to see if it confirm with what the component does now, especially when it comes to the group-on and sort-on attributes as these are involved.

PR #54, opened 2017-09-28
Allow a default column width value which is dependent on the editable property

JaySunSyn:
The optimal default width does vary depending on the `editable` property.
Currently needed in column datetime.

Added properties
- cellWidth
- editCellWith

Removed property
- width

### DIFF
--- a/cosmoz-omnitable-column-date-behavior.html
+++ b/cosmoz-omnitable-column-date-behavior.html
@@ -96,7 +96,7 @@
 				type: String,
 				computed: '_computeTooltip(title, _filterText)'
 			},
-			width: {
+			cellWidth: {
 				type: String,
 				value: '180px'
 			}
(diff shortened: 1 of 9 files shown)


cosmoz-omnitable-column-datetime.html
```diff
@@ -56,6 +56,16 @@ <h3 style="margin: 0;">[[ title ]]</h3>
...
 				},
 
+				editableWidth: {
+					type: String,
+					value: '300px'
+				},
+
+				cellWidth: {
+					type: String,
+					value: '180px'
```
plequang: Seems 180px is not enough. In full-demo, the date-time column is overflowing.

cosmoz-omnitable-column-date-behavior.html
```diff
@@ -92,10 +92,25 @@
...
+
 			_tooltip: {
 				type: String,
 				computed: '_computeTooltip(title, _filterText)'
 			},
+
+			cellWidth: {
+				type: String,
+				value: '180px'
+			},
```
plequang: Edit cell width need to be increased for dates, otherwise it will always overflow

Issue #57, opened 2017-10-10
Use DateTimeFormat to render dates and times


nomego:
Our -date, -time and -datetime columns should use Intl.DateTimeFormat to render cell values.
Similar to the -number column, this will honor browser language/locale settings.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat

For now, we should still use moment for other kind of rendering (timeago) and comparison (valid/before/after) functions.

PR #59, opened 2017-10-12
Uses datetimeformat and the browsers language to determine date forma…

JockeCK:
…t. #57 

Not done and not that happy with initial solution either. But need some feedback on direction to take.

Have tested on our demos and seem to work for Swedish/american date standards in desktop chrome at least.


cosmoz-omnitable-column-date-behavior.html
```diff
@@ -231,19 +231,29 @@
 		},
 
 		renderDateValue(item, valuePath) {
```
Wurper: Call getString(item, valuePath) to remove duplicate code.

cosmoz-omnitable-column-datetime.html
```diff
@@ -98,7 +98,8 @@ <h3 style="margin: 0;">[[ title ]]</h3>
 				if (!mom || !mom.isValid()) {
 					return;
 				}
-				return mom.format('YYYY-MM-DDTHH:mm:ss');
+				return this.formatter.format(mom);
+				//return mom.format('YYYY-MM-DDTHH:mm:ss');
```
Wurper: Remove commented row

cosmoz-omnitable-column-date-behavior.html
```diff
@@ -231,19 +231,29 @@
...
-			var value = this.get(valuePath, item);
+			var value = this.get(valuePath, item),
+				month;
 			if (!value) {
 				return;
 			}
-			return moment(value).format(this._inputFormat);
+			if (value.length === 10) {
```
nomego: First of all, check if `value` is a `Date` already, if not, just use `new Date(string)` or `Date.parse(dateString)` to cover more formats.

cosmoz-omnitable-column-date.html
```diff
@@ -4,14 +4,15 @@
...
+
 <link rel="import" href="../cosmoz-i18next/cosmoz-i18next.html">
 
 <link rel="import" href="cosmoz-omnitable-column-date-behavior.html">
 
 <dom-module id="cosmoz-omnitable-column-date">
 	<template>
 		<template class="cell">
-			[[ renderDateValue(item, valuePath, userFormat, locale) ]]
+			[[ renderDateValue(item, valuePath) ]]
```
nomego: This needs to observe the `formatter` in case locale or options change.

cosmoz-omnitable-column-date-behavior.html
```diff
@@ -116,13 +107,22 @@
...
-
-			_inputFormat: {
+			formatter: {
+				type: Object,
+				computed: '_computeFormatter(locale)'
+			},
+			locale: {
 				type: String,
-				value: 'YYYY-MM-DD'
+				value: function () {
```
nomego: Rather have the default value as `null` to make the browser implementation select the locale.
If it's `undefined`, the formatter will never be computed, but if it's `null`, the formatter will be instantiated with an `undefined` argument for locale, making the browser select the best fit.
`locale` should only be set/used when the component property is set manually.

cosmoz-omnitable-column-date-behavior.html
```diff
@@ -231,19 +231,29 @@
...
+				month;
 			if (!value) {
 				return;
 			}
-			return moment(value).format(this._inputFormat);
+			if (value.length === 10) {
+				month = parseInt(value.substring(5, 7), 10) - 1;
+				value = Date.UTC(value.substring(0, 4), month, value.substring(8));
+			}
+			return this.formatter.format(value);
```
nomego: As described above, make sure this method outputs `yyyy-mm-dd` - don't use the same formatter.

cosmoz-omnitable-column-date-behavior.html
```diff
@@ -231,19 +231,29 @@
...
 
 		getString(item, valuePath) {
-			var value = this.get(valuePath || this.valuePath, item);
+			var value = this.get(valuePath || this.valuePath, item),
+				month;
 			if (!value) {
 				return '';
 			}
-			return moment(value).format(this.userFormat);
+			if (value.length === 10) {
```
nomego: Same as above

cosmoz-omnitable-column-datetime.html
```diff
@@ -98,7 +98,8 @@ <h3 style="margin: 0;">[[ title ]]</h3>
 				if (!mom || !mom.isValid()) {
 					return;
 				}
-				return mom.format('YYYY-MM-DDTHH:mm:ss');
+				return this.formatter.format(mom);
```
nomego: mom is a moment instance, shouldn't be compatible?
Also, `<input type="datetime-local">` has a similar requirement as date - value should be in the `YYYY-MM-DDTHH:mm` format -
 see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local

cosmoz-omnitable-column-date-behavior.html
```diff
@@ -5,6 +5,15 @@
 	Cosmoz.DateColumnBehavior = {
 		properties: {
 			/**
+			* This is the format used to interact with the user, i.e. the format used
+			* to render dates and to parse date entered by the user.
+			* Should be a moment.js date format (http://momentjs.com/docs/#/displaying/format/)
+			*/
+			userFormat: {
```
nomego: Removing this was the right thing to do since this will be decided by browser, no need to re-add it.

cosmoz-omnitable-column-date-behavior.html
```diff
@@ -196,13 +202,13 @@
 			var filter = filterNotify.base,
 				text = [];
 			if (filter.start !== null) {
-				text.push(filter.start.locale(locale).format(this.userFormat));
+				text.push(this.formatter.format(filter.start.i));
 			}
-			text.push(' - ');
```
nomego: This was intended to be here to indicate a period without end, removing it will be interpreted as a specific date, which is not true.

cosmoz-omnitable-column-date-behavior.html
```diff
@@ -196,13 +202,13 @@
...
 			if (filter.end !== null) {
-				text.push(filter.end.locale(locale).format(this.userFormat));
+				text.push(' - ');
+				text.push(this.formatter.format(filter.start.i));
```
nomego: Copy/paste typo, test more

cosmoz-omnitable-column-date-behavior.html
```diff
@@ -230,30 +235,23 @@
...
+			var value = this.get(valuePath, item);
 			if (!value) {
 				return;
 			}
-			if (value.length === 10) {
-				month = parseInt(value.substring(5, 7), 10) - 1;
-				value = Date.UTC(value.substring(0, 4), month, value.substring(8));
-			}
-			return this.formatter.format(value);
+			return value.toISOString().slice(0, 10);
```
nomego: Make sure `value` is a valid `Date` before calling `toISOString()`

cosmoz-omnitable-column-date-behavior.html
```diff
@@ -230,30 +235,23 @@
...
-			if (!value) {
+		getString(item, valuePath, formatter = this.formatter) {
+			var value = this.get(valuePath || this.valuePath, item);
+			if (!value || !formatter) {
 				return '';
 			}
-			if (value.length === 10) {
-				month = parseInt(value.substring(5, 7), 10) - 1;
-				value = Date.UTC(value.substring(0, 4), month, value.substring(8));
+			if (Object.prototype.toString.call(value) !== '[object Date]') {
```
nomego: Did you test whether this checks for a valid date? My tests say it doesn't.
Also this check needs to be split up in two - one to make sure it's a `Date` object and one to check the validity.
If the last one fails, return something to indicate this.. maybe `toString()` on the `Date` object, which should be `"Invalid Date"`.

So the logic should be something like:

```js
if (!(value instanceof Date)) {
  value = new Date(value); // or maybe even new Date(Date.parse(value));
}
if (isNaN(value)) {
  return 'Invalid Date'; // or value.toString();
}
```

cosmoz-omnitable-column-datetime.html
```diff
@@ -13,12 +13,12 @@
...
 	<template>
 		<template class="cell">
-			[[ renderDateValue(item, valuePath, userFormat, locale) ]]
+			[[ getString(item, valuePath, formatter) ]]
 		</template>
 
 		<template class="edit-cell">
 			<datetime-local no-label-float type="datetime-local" on-change="_dateValueChanged"
-				value="[[ renderDateValue(item, valuePath) ]]">
+				value="[[ renderISODateValue(item, valuePath, locale) ]]">
```
nomego: ISO Date should be the same regardless of locale

cosmoz-omnitable-column-datetime.html
```diff
@@ -98,8 +98,7 @@ <h3 style="margin: 0;">[[ title ]]</h3>
 				if (!mom || !mom.isValid()) {
 					return;
 				}
-				return this.formatter.format(mom);
-				//return mom.format('YYYY-MM-DDTHH:mm:ss');
+				return mom.format('YYYY-MM-DDTHH:mm:ss');
```
nomego: We should #usetheplatform and rather do a `toISOString().slice(0,19)` on a Date object.

cosmoz-omnitable-column-datetime.html
```diff
@@ -92,14 +81,27 @@ <h3 style="margin: 0;">[[ title ]]</h3>
 
 			/** OVERRIDES */
 
-			/** cosmoz-omnitable-column-date-behavior */
+			/** cosmoz-omnitable-column-date-behavior*/
 
 			_formatISODate(mom) {
-				if (!mom || !mom.isValid()) {
```
nomego: Drop comments completely.
Also `mom` should be renamed since it's not moment anymore.
Can we guarantee that the passed argument is a (valid) date?

cosmoz-omnitable-column-datetime.html
```diff
@@ -92,14 +81,27 @@ <h3 style="margin: 0;">[[ title ]]</h3>
...
+				//if (!mom || !mom.isValid()) {
+				if (!mom) {
 					return;
 				}
-				return mom.format('YYYY-MM-DDTHH:mm:ss');
+				return this.formatter.format(mom);
 			},
+
+			/** OVERRIDES */
```
nomego: I don't think this stuff is necessary to repeat, it's more of header to describe things that will override something.

cosmoz-omnitable-column-datetime.html
```diff
@@ -46,15 +45,6 @@ <h3 style="margin: 0;">[[ title ]]</h3>
 					type: Number,
 					computed: '_computeStep(userFormat)'
 				},
-				/**
-				 * This is the format used to interact with the user, i.e. the format used
-				 * to render dates and to parse date entered by the user.
-				 * Should be a moment.js date format (http://momentjs.com/docs/#/displaying/format/)
-				 */
-				userFormat: {
```
nomego: `userFormat` is still referenced in `_computeStep()` - needs to be refactored too

cosmoz-omnitable-column-time.html
```diff
@@ -48,7 +46,7 @@ <h3 style="margin: 0;">[[ title ]]</h3>
 			properties: {
 				userFormat: {
 					type: String,
-					value: 'HH:mm:ss.SSS'
+					value: 'YYYY-MM-DDTHH:mm:ss.SSS'
```
nomego: `userFormat` should be dropped completely after the last place is refactored (`_computeStep()`) as with datetime

cosmoz-omnitable-column-time.html
```diff
@@ -60,17 +58,51 @@ <h3 style="margin: 0;">[[ title ]]</h3>
 					computed: '_computeStep(userFormat)'
 				},
 
+				_filterInput: {
```
nomego: This `_filterInput`, `_filterText`, `_tooltip`, `_computeTooltip` and `filter` properties should already be inherited from date behavior?

cosmoz-omnitable-column-time.html
```diff
@@ -60,17 +58,51 @@ <h3 style="margin: 0;">[[ title ]]</h3>
...
+					computed: '_computeTooltip(title, _filterText)'
 				}
 			},
 
+			_computeTimeFormatter(locale) {
+				var timeFormatOption = {
+					hour: 'numeric',
+					minute: 'numeric',
+					second: 'numeric',
+					hour12: false
```
nomego: By default `hour12` is controlled by locale and we shouldn't override it.

cosmoz-omnitable-column-time.html
```diff
@@ -60,17 +58,51 @@ <h3 style="margin: 0;">[[ title ]]</h3>
...
+					type: Object,
+					computed: '_computeFilter(_filterInput.start, _filterInput.end)'
+				},
+
 				_placeholderDate: {
 					type: String,
 					value: '1970-01-01'
 				},
+		
+				timeFormatter: {
```
nomego: Why can't we only override the `formatter` inherited from date behavior?
We can probably reuse more methods then?

cosmoz-omnitable-column-time.html
```diff
@@ -81,82 +113,91 @@ <h3 style="margin: 0;">[[ title ]]</h3>
...
 			},
+		
+			/**OVERRIDES */
 
-			_computePlaceholderDateTimeFormat(userFormat) {
-				return 'YYYY-MM-DD ' + userFormat;
-			},
+			/** cosmoz-omnitable-column-date-behavior */
 
+			_computeFilterText(locale, filterNotify) {
```
nomego: If we override `formatter` we shouldn't need to override this.
If we don't, this should probably have the same requirements as the method from date behavior (default formatter, checks, etc)

cosmoz-omnitable-column-time.html
```diff
@@ -81,82 +113,91 @@ <h3 style="margin: 0;">[[ title ]]</h3>
...
+				return text.length > 1 ? text.join('') : null;
+			},
 			_timeValueChanged(e) {
 				var
 					input = e.target,
 					timeString = input.value,
 					item = e.model.item,
-					oldDateMom = moment(item.date),
-					newDateMom = moment(oldDateMom.format('YYYY-MM-DD') + ` ${timeString} ` + oldDateMom.format('Z')),
+					oldDateMom = this.getDate(item.date),
```
nomego: In the old implementation, both new and old dates were moment, now only old date is Date, new Date is string?
Also formatFn converted date -> string, now it converts string -> date ?
formatFn should rather use the formatter and newDate should be a Date.
Also all "mom"-references should be dropped since we're not using moment.

cosmoz-omnitable-column-time.html
```diff
@@ -81,82 +113,91 @@ <h3 style="margin: 0;">[[ title ]]</h3>
...
-			/** cosmoz-omnitable-column-date-behavior */
 
 			_computeFilter(startTime, endTime) {
-				var startMoment = startTime ? moment(this._placeholderDate + ' ' + startTime) : null,
-					endMoment = endTime ? moment(this._placeholderDate + ' ' + endTime) : null;
-
 				return {
-					start: startMoment,
-					end: endMoment
+					start: startTime ? startTime : null,
```
nomego: This can be rewritten `startTime || null`

cosmoz-omnitable-column-time.html
```diff
@@ -81,82 +113,91 @@ <h3 style="margin: 0;">[[ title ]]</h3>
...
 
-			_applySingleFilter(filter, item) {
+			_applySingleFilter(filter, item, timeFormatter = this.timeFormatter) {
 				var value = this.get(this.valuePath, item),
-					timeString = moment(value).format(this.userFormat),
-					mom = moment(this._placeholderDate + ' ' + timeString, this._placeholderDateTimeFormat);
+					dateToCheck = this.getDate(value);
 
-				if (filter.start && mom.isBefore(filter.start)) {
+				if (filter.start && timeFormatter.format(dateToCheck) < filter.start) {
```
nomego: Dates can be compared directly. Isn't filter.start a date?

cosmoz-omnitable-column-time.html
```diff
@@ -81,82 +113,91 @@ <h3 style="margin: 0;">[[ title ]]</h3>
...
-					timeString = moment(value).format(this.userFormat),
-					mom = moment(this._placeholderDate + ' ' + timeString, this._placeholderDateTimeFormat),
-					compareValue = mom.unix(),
-					milliFormatMatch = this.userFormat.match('S+');
-
-				if (milliFormatMatch) {
-					compareValue = parseInt(mom.unix().toString() + mom.format(milliFormatMatch[0]), 10);
-				}
-				return compareValue;
+					numericValue = new Date(value);
```
nomego: `const date = `

cosmoz-omnitable-column-time.html
```diff
@@ -81,82 +113,91 @@ <h3 style="margin: 0;">[[ title ]]</h3>
...
+				return numericValue.getTime();
 			},
-
-			getLocaleString(item, valuePath, userFormat, locale) {
-				var value = this.get(valuePath, item);
-				if (!value) {
+		
+			getLocaleString(item, valuePath, timeFormatter = this.timeFormatter) {
```
nomego: use `valuePath = this.valuePath` to avoid checks withing the method
This method copies a lot from getDate, reuse it

cosmoz-omnitable-column-date-behavior.html
```diff
@@ -116,108 +107,116 @@
...
-				second = moment(secondRaw);
+			const doMax = max === 1,
+				first = this.getDate(firstRaw),
+				second = this.getDate(secondRaw);
+			if (!first && !second) {
+				return null;
+			}
 
-			if (!first.isValid() && second.isValid()) {
+			if (!this._isValidDate(first) && second) {
```
nomego: This actually doesn't check that `second` is valid.
I think this whole method would be safer and more readable by adding `firstValid = this._isValidDate(first)` for both and having a pattern like
`if (!validFirst && validSecond) { return second; }`

cosmoz-omnitable-column-date-behavior.html
```diff
@@ -116,108 +107,116 @@
...
+				return doMax ? second : first;
 			}
 
-			return doMax ? second : first;
+			return doMax ? first : second;
 		},
 
 		_computePossibleDateRange(change) {
-			var values = change.base;
+			let values = change.base;
```
nomego: `values` reference never change, use `const`

cosmoz-omnitable-column-date-behavior.html
```diff
@@ -116,108 +107,116 @@
...
 				return {
 					min: null,
 					max: null
 				};
 			}
-			return values.reduce(function (acc, item) {
-				var mom = moment(item);
-				if (!acc.min || mom.isBefore(acc.min)) {
-					acc.min = mom;
+			return values.reduce((acc, item) => {
```
nomego: Not sure if this is what we discussed in the office, and not sure this change is the right way, but it _could_ be rewritten as:
```js
values
  .map(item => this.getDate(item))
  .filter(item => this._isValidDate(item))
  .reduce((acc, item) => {
    // current reduce but without mapping and checks
  })

File: cosmoz-omnitable-column-date-behavior.html [hunk 32]
```diff
@@ -116,108 +107,116 @@
...
 			};
 		},
 
-		_computeFilterText(locale, filterNotify) {
-			var filter = filterNotify.base,
-				text = [];
-			if (filter.start !== null) {
-				text.push(filter.start.locale(locale).format(this.userFormat));
+		_computeFilterText(locale, filterNotify, formatter = this.formatter) {
+			let text = [];
```
nomego: const

cosmoz-omnitable-column-date-behavior.html
```diff
@@ -116,108 +107,116 @@
...
+			const startDate = filterNotify.base.start,
+				endDate = filterNotify.base.end;
+			if (!formatter) {
+				return null;
+			}
+			if (startDate && startDate !== null) {
```
nomego: The truthy check on `startDate` also includes `startDate !== null` so it can be removed.
Although, isn't `startDate` a date here already? It would be better to use `_isValidDate()`

cosmoz-omnitable-column-date-behavior.html
```diff
@@ -116,108 +107,116 @@
...
+				text.push(formatter.format(this.getDate(startDate)));
 			}
 			text.push(' - ');
-			if (filter.end !== null) {
-				text.push(filter.end.locale(locale).format(this.userFormat));
+			if (endDate && endDate !== null) {
```
nomego: Same as `startDate`

cosmoz-omnitable-column-date-behavior.html
```diff
@@ -116,108 +107,116 @@
...
 				formatFn = (value) => {
-					return moment(value).locale(this.locale).format(this.userFormat);
-				},
-				dateValue;
-
+					return this.formatter.format(value);
+				};
 			try {
-				dateValue = new Date(value);
+				let dateValue = new Date(value);
```
nomego: You should use getDate + isValidDate instead of try/catch

cosmoz-omnitable-column-date-behavior.html
```diff
@@ -226,24 +225,43 @@
...
+				return null;
+			}
+			return numericValue.getTime();
 		},
 
-		renderDateValue(item, valuePath) {
-			var value = this.get(valuePath, item);
+		renderISODateValue(item, valuePath) {
+			const value = this.get(valuePath, item);
 			if (!value) {
```
nomego: use isValidDate

cosmoz-omnitable-column-date-behavior.html
```diff
@@ -226,24 +225,43 @@
...
 				return;
 			}
-			return moment(value).format(this._inputFormat);
+			return value.toISOString().slice(0, 10);
 		},
 
-		getString(item, valuePath) {
-			var value = this.get(valuePath || this.valuePath, item);
-			if (!value) {
+		getString(item, valuePath, formatter = this.formatter) {
```
nomego: use `valuePath = this.valuePath` instead of `|| this.valuePath` below

cosmoz-omnitable-column-date-behavior.html
```diff
@@ -226,24 +225,43 @@
...
+			if (!dateValue) {
 				return '';
 			}
-			return moment(value).format(this.userFormat);
+			if (!(dateValue instanceof Date)) {
+				dateValue = new Date(dateValue);
+			} else {
+				return dateValue;
+			}
+			if (isNaN(dateValue)) {
```
nomego: This should be moved to _isValidDate()

cosmoz-omnitable-column-date-behavior.html
```diff
@@ -116,108 +107,116 @@
...
-				text.push(filter.start.locale(locale).format(this.userFormat));
+		_computeFilterText(locale, filterNotify, formatter = this.formatter) {
+			let text = [];
+			const startDate = filterNotify.base.start,
+				endDate = filterNotify.base.end;
+			if (!formatter) {
+				return null;
+			}
+			if (startDate && startDate !== null) {
+				text.push(formatter.format(this.getDate(startDate)));
```
nomego: getDate() shouldn't be needed or should run earlier if needed

cosmoz-omnitable-column-date-behavior.html
```diff
@@ -254,17 +272,22 @@
...
 
-			if (filter.end && mom.isAfter(filter.end)) {
+			if (filter.end && dateToCheck > filter.end) {
+				return false;
+			}
+			return true;
+		},
+	
+		_isValidDate(value) {
+			if (!value || !value === null || value === 'Invalid Date') {
```
nomego: `!value === null` will never be true :) Also, move the isNaN check here instead of the string comparison

cosmoz-omnitable-column-datetime.html
```diff
@@ -65,41 +66,63 @@ <h3 style="margin: 0;">[[ title ]]</h3>
...
+		
+			_computeFilterText(locale, filterNotify, formatter = this.formatter) {
+				let text = [];
+				const startTime = filterNotify.base.start,
+					endTime = filterNotify.base.end;
+				if (!formatter) {
+					return null;
+				}
+				if (startTime && startTime !== null) {
+					text.push(this._formatISODate(startTime));
```
nomego: This should rather use the formatter since it will be a tooltip for the user.
And if you switch to the formatter I think you can drop this implementation and use the one from date behavior ?

cosmoz-omnitable-column-datetime.html
```diff
@@ -1,24 +1,20 @@
...
 	<template>
 		<template class="cell">
-			[[ renderDateValue(item, valuePath, userFormat, locale) ]]
+			[[ getString(item, valuePath, formatter) ]]
 		</template>
 
 		<template class="edit-cell">
 			<datetime-local no-label-float type="datetime-local" on-change="_dateValueChanged"
-				value="[[ renderDateValue(item, valuePath) ]]">
+				value="[[ renderISODateValue(item, valuePath) ]]">
```
nomego: This should probably be renamed `renderISODateTimeValue` to be less confusing

cosmoz-omnitable-column-datetime.html
```diff
@@ -28,10 +24,10 @@
 				<div class="dropdown-content" slot="dropdown-content" style="padding: 15px; min-width: 100px;">
 					<h3 style="margin: 0;">[[ title ]]</h3>
 					<datetime-local min="[[ _formatISODate(_fromMin) ]]" max="[[ _formatISODate(_fromMax) ]]"
```
nomego: `min`/`max` should rather use `renderISODateTimeValue` or even a method that returns the first 16 characters of the ISOString

cosmoz-omnitable-column-datetime.html
```diff
@@ -28,10 +24,10 @@
...
-						date-label="[[ _('From date', t) ]]" time-label="[[ _('From time', t) ]]" value="{{ _filterInput.start }}" step$="[[_filterStep]]">
+						date-label="[[ _('From date', t) ]]" time-label="[[ _('From time', t) ]]" value="{{ _filterInput.start }}" step$="[[filterStep]]">
 					</datetime-local>
 					<datetime-local on-datetime-local-value-changed="_valueChanged" min="[[ _formatISODate(_toMin) ]]" max="[[ _formatISODate(_toMax) ]]"
```
nomego: Same as above

cosmoz-omnitable-column-datetime.html
```diff
@@ -65,41 +66,63 @@ <h3 style="margin: 0;">[[ title ]]</h3>
...
+					text.push(this._formatISODate(endTime));
+				}
+				return text.length > 1 ? text.join('') : null;
+			},
+
+			_formatISODate(filterValue) {
+				if (!this._isValidDate(filterValue)) {
+					return '';
+				}
+				return this.formatter.format(filterValue);
```
nomego: This is not guaranteed to format an ISO date.
Not sure if this is needed after above comments?

cosmoz-omnitable-column-time.html
```diff
@@ -39,124 +38,113 @@ <h3 style="margin: 0;">[[ title ]]</h3>
...
 			_timeValueChanged(e) {
-				var
-					input = e.target,
+				const input = e.target,
 					timeString = input.value,
 					item = e.model.item,
-					oldDateMom = moment(item.date),
-					newDateMom = moment(oldDateMom.format('YYYY-MM-DD') + ` ${timeString} ` + oldDateMom.format('Z')),
+					oldTime = this.getDate(item.date),
+					newTime = oldTime.toISOString().slice(0, 10) + ` ${timeString} `,
```
nomego: Still not a Date, shouldn't it be?

cosmoz-omnitable-column-time.html
```diff
@@ -39,124 +38,113 @@ <h3 style="margin: 0;">[[ title ]]</h3>
...
 					formatFn = (value) => {
-						return moment(value).locale(this.locale).format(this.userFormat);
+						return value;
```
nomego: Is this wrapper function even needed now?

cosmoz-omnitable-column-time.html
```diff
@@ -39,124 +38,113 @@ <h3 style="margin: 0;">[[ title ]]</h3>
...
 					};
-
 				try {
-					this.set(this.valuePath, newDateMom.toDate(), item);
+					this.set(this.valuePath, newTime, item);
```
nomego: Old code seems to expect a Date here

cosmoz-omnitable-column-time.html
```diff
@@ -39,124 +38,113 @@ <h3 style="margin: 0;">[[ title ]]</h3>
...
 				} catch (err) {
 					input.errorMessage = 'Invalid date';
 				}
-				this._fireItemChangeEvent(item, this.valuePath, oldDateMom.toDate(), formatFn.bind(this));
+				this._fireItemChangeEvent(item, this.valuePath, formatFn.bind(this));
```
nomego: Missing the third (date) argument

cosmoz-omnitable-column-time.html
```diff
@@ -39,124 +38,113 @@ <h3 style="margin: 0;">[[ title ]]</h3>
...
+			},
+		
+			getLocaleString(item, valuePath, formatter = this.formatter) {
```
nomego: use getDate, isValidDate here.
If renamed, can't Date behavior getString be used?

cosmoz-omnitable-column-time.html
```diff
@@ -9,12 +8,12 @@
 <dom-module id="cosmoz-omnitable-column-time">
 	<template>
 		<template class="cell">
-			[[ getLocaleString(item, valuePath, userFormat, locale) ]]
+			[[ getLocaleString(item, valuePath, formatter) ]]
```
nomego: Should probably use `getString()` like the other implmentations (just rename)

cosmoz-omnitable-column-time.html
```diff
@@ -39,124 +38,113 @@ <h3 style="margin: 0;">[[ title ]]</h3>
...
+				const value = this.get(this.valuePath, item),
+					dateToCheck = this.getDate(value);
 
-			_applySingleFilter(filter, item) {
-				var value = this.get(this.valuePath, item),
-					timeString = moment(value).format(this.userFormat),
-					mom = moment(this._placeholderDate + ' ' + timeString, this._placeholderDateTimeFormat);
-
-				if (filter.start && mom.isBefore(filter.start)) {
+				if (filter.start && formatter.format(dateToCheck) < filter.start) {
```
nomego: Probably typo?
Compare dates directly, formatter will return an unpredictable string.
Same next if
Can't behavior filter be inherited?

cosmoz-omnitable-column-time.html
```diff
@@ -39,124 +38,113 @@ <h3 style="margin: 0;">[[ title ]]</h3>
...
+		
+			_computeFilterText(locale, filterNotify, formatter = this.formatter) {
+				let text = [];
+				const startTime = filterNotify.base.start,
+					endTime = filterNotify.base.end;
+				if (!formatter) {
+					return null;
+				}
+				if (startTime && startTime !== null) {
+					text.push(startTime);
```
nomego: Text should be formatted

cosmoz-omnitable-column-time.html
```diff
@@ -39,124 +38,113 @@ <h3 style="margin: 0;">[[ title ]]</h3>
...
+					end: endTime || null
+				};
 			},
-
-			getLocaleString(item, valuePath, userFormat, locale) {
-				var value = this.get(valuePath, item);
-				if (!value) {
-					return '';
+		
+			_computeFilterText(locale, filterNotify, formatter = this.formatter) {
```
nomego: This could probably be inherited?

cosmoz-omnitable-column-time.html
```diff
@@ -39,124 +38,166 @@ <h3 style="margin: 0;">[[ title ]]</h3>
...
+				const time = this.getTime(this.get(valuePath, item)),
+					date = new Date(this._fixedDate + 'T' + time);
+				if (!this._isValidDate(date)) {
+					return null;
+				}
+				return date.getTime();
+			},
 
-			/** cosmoz-omnitable-column-date-behavior */
+			getTimeString(item, valuePath = this.valuePath) {
```
nomego: This should override `getString()` since it's needed for CSV and possibly XLS export

cosmoz-omnitable-column-time.html
```diff
@@ -39,124 +38,166 @@ <h3 style="margin: 0;">[[ title ]]</h3>
...
-					computed: '_computePlaceholderDateTimeFormat(userFormat)'
-				}
-			},
-
-			/** ELEMENT LIFECYCLE */
+					computed: '_computeFilterText(locale, filter.*)'
+				},
 
-			/** ELEMENT BEHAVIOR */
+				filter: {
```
nomego: Same declaration as date behavior, can be removed

cosmoz-omnitable-column-date-behavior.html
```diff
@@ -60,7 +51,7 @@
 
 			_filterText: {
 				type: String,
-				computed: '_computeFilterText(locale, filter.*)'
+				computed: '_computeFilterText(filter.*)'
```
nomego: This needs to observe `formatter` as well

cosmoz-omnitable-column-date-behavior.html
```diff
@@ -116,134 +107,152 @@
...
-			try {
-				dateValue = new Date(value);
+				formatFn = value => this.formatter.format(value),
+				dateValue = this.getDate(value);
+			if (this._isValidDate(dateValue)) {
 				this.set(this.valuePath, dateValue, item);
-			} catch (err) {
-				input.errorMessage = 'Invalid date';
 			}
 			this._fireItemChangeEvent(item, this.valuePath, oldValue, formatFn.bind(this));
```
nomego: I assume we shouldn't fire these for invalid dates. We should probably `return` on invalid dates,

```js
if (!this._isValidDate(dateValue)) {
  return;
}
this.set(...)
this._fire..()
```

cosmoz-omnitable-column-date-behavior.html
```diff
@@ -116,134 +107,152 @@
...
+				return null;
+			}
+			return numericValue.getTime();
 		},
 
-		renderDateValue(item, valuePath) {
-			var value = this.get(valuePath, item);
-			if (!value) {
+		renderISODateValue(item, valuePath) {
+			const value = this.get(valuePath, item);
```
nomego: I think we need to run `getDate()` on this as well since we're not sure what we might expect in the data.

cosmoz-omnitable-column-date-behavior.html
```diff
@@ -254,17 +263,29 @@
 		},
 
 		_applySingleFilter(filter, item) {
-			var value = this.get(this.valuePath, item),
-				mom = moment(value);
+			const dateToCheck = this.getDate(this.get(this.valuePath, item));
 
-			if (filter.start && mom.isBefore(filter.start)) {
+			if (filter.start && dateToCheck < filter.start) {
```
nomego: We probably want to check date validity here as well.
`filter.start && !this._isValidDate(dateToCheck) || dateToCheck < filter.start)` ..
We should only filter invalid dates out if a filter has been set though.

Same for `filter.end` below

cosmoz-omnitable-column-time.html
```diff
@@ -39,124 +38,147 @@ <h3 style="margin: 0;">[[ title ]]</h3>
...
-					};
-
-				try {
-					this.set(this.valuePath, newDateMom.toDate(), item);
-				} catch (err) {
-					input.errorMessage = 'Invalid date';
+					oldTime = this.getDate(item.date),
+					newTime = this.getDate(oldTime.toISOString().slice(0, 10) + ` ${timeString} `),
+					formatFn = value => value;
+				if (this._isValidDate(newTime)) {
```
nomego: Probably want to do the same here as above, if not valid, return, if valid set + fire.

PR #63, opened 2017-10-16
XLSX export/download

### LINKED ISSUE #8: XLSX export/download
Implement a similar action like the current CSV export action, but use https://github.com/Neovici/nullxlsx/ to export/download XLSX.

programmer4web:
Added XLSX export/download

fixes #8

### DIFF
--- a/cosmoz-omnitable.js
+++ b/cosmoz-omnitable.js
@@ -1,4 +1,4 @@
-/*global Cosmoz, Polymer, window, saveAs */
+/*global Cosmoz, Polymer, window, saveAs, NullXlsx */
 (function () {
 
 	'use strict';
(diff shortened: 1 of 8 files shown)


import-nullxlsx.html
```diff
@@ -0,0 +1 @@
+<script src="../nullxlsx/dist/nullxlsx.min.js"></script>
```
nomego: Wouldn't it make more sense to have this in the nullxlsx repo?

cosmoz-omnitable.js
```diff
@@ -860,6 +876,40 @@
 			}));
 		},
 
+		/**
+		 * Makes the data ready to be exported as XLSX.
+		 * @returns {Array} data Array of prepared rows.
+		 */
+		_prepareXlsxData: function () {
```
nomego: The biggest advantage of saving as XLSX over CSV is the ability to describe the column types, which this doesn't seem to do.
Not having tested the code, I would assume that Excel would have similar difficulties as with CSV to figure out what is numbers, dates, money, etc without explicit column types, but I might be wrong?

programmer4web: At the moment it doesn't handle data types as String is being used.
It seems that the library supports data types so I will try to export it with correct data type.

cosmoz-omnitable-column-date-behavior.html
```diff
@@ -246,6 +246,11 @@
 			return moment(value).format(this.userFormat);
 		},
 
+		toXlsxValue: function () {
```
megheaiulian: @programmer4web Revert change to `OmnitableDateBehavior` because `moment` will be removed in  #59.

PR #64, opened 2017-10-16
Persist filtering in URL

### LINKED ISSUE #61: Persist filtering in URL (hashhash)
After applying filters to data sets in omnitable and then moving to another view, users expect to be able to go back to the omnitable view, maintaining the filters configured.

To support this, we should manage a state in the URL "hashhash" part, much like cosmoz-tabs does today to remember open/active tab.
This will also have the added benefit of users being able to send each other links to views with omnitables and predefined filters.
It will also enable a clean solution to link from stats to a view with filters to match the stats.

If possible, we should share/reuse the approach from cosmoz-tabs, and it might mean doing a new small component to offer the functionality.

As far as view-implementation goes, we have to (like tabs) consider that multiple omnitables are present in the same view.
Omnitable columns have a `name` attribute that should be unique within the omnitable so we should use this to specify which filter in the omnitable that we refer to.

An initial idea is to have a known prefix (`ot` or `omnitable` or `omnitable-filter` or `cosmoz-omnitable`) followed by an (optional) omnitable ID for the view in case there are multiple, followed by the column name, `=`, filter.

For example:
`omnitable-filter-status=open` for a single omnitable in a view
`omnitable-filter-ot1-currency=SEK,USD` for named omnitable (with `id=`?) and multi-select

We also have to make sure that the filters are applied at the right time so they don't get overwritten by computes, late loads, etc.

Did I miss anything? Pitfalls, gotchas, features?

megheaiulian:
Persist filtering in URL:

* Added groupOn, sortOn, descending to hash-param
* Added demo with cosmoz-page-router

fixes #61

### DIFF
--- a/cosmoz-omnitable.js
+++ b/cosmoz-omnitable.js
@@ -199,8 +199,7 @@
 		observers: [
 			'_dataChanged(data.*)',
 			'_debounceSortItems(sortOn, descending, filteredGroupedItems)',
-			'_routeHashParamsChanged(_routeHashParams.*, hashParam)',
-			'_routeHashParamsForFiltersChanged(_routeHashParams.*, hashParam, columns)',
+			'_routeHashParamsChanged(_routeHashParams.*, hashParam, columns)',
 			'_paramForRouteChanged( hashParam, "groupOn", groupOn)',
 			'_paramForRouteChanged( hashParam, "sortOn", sortOn)',
 			'_paramForRouteChanged( hashParam, "descending", descending)'
@@ -1031,28 +1030,30 @@
 			gl.highlightItem(i, reverse);
 		},
 
-		_routeHashParamsChanged: function (changes, hashParam) {
-			if (!changes || !hashParam) {
+		_routeHashParamsChanged: function (changes, hashParam, columns) {
+			if (!changes || !hashParam || !columns || !columns.length) {
 				return;
 			}
+
 			PROPERTY_HASH_PARAMS.forEach(key => {
-				var hashValue =  this.get(['_routeHashParams', hashParam + '-' + key]),
+				let hashValue =  this.get(['_routeHashParams', hashParam + '-' + key]),
 					deserialized = this.deserialize(hashValue, this.properties[key].type);
 
-				if (hashValue !== undefined && deserialized !== this.get(key)) {
-					this.set(key, deserialized);
+				if (hashValue === undefined ||  deserialized === this.get(key)) {
+					return;
 				}
-			});
-		},
 
-		_routeHashParamsForFiltersChanged: function (changes, hashParam, columns) {
-			if (!changes || !hashParam || !(columns && columns.length)) {
-				return;
-			}
-			Object.keys(changes.base).map(key => {
-				var matches = key.match(/^full\-filter\-\-([a-z0-9\-]+)$/),
+				this.set(key, deserialized);
+				console.log('route changed for', key, deserialized);
+			});
+
+			let rule = new RegExp('^' + hashParam + '-filter\-\-([a-z0-9\-]+)$'),
+				routeParams = changes.base;
+
+			Object.keys(routeParams).forEach(key => {
+				let matches = key.match(rule),
 					name = matches && matches[1],
-					hashValue,
+					hashValue = routeParams[key],
 					column,
 					value;
 
@@ -1065,14 +1066,16 @@
 					return;
 				}
 
-				hashValue = changes.base[key];
 				value = this.serialize(column.filter, Object);
 
-				if (hashValue !== value) {
-					column.set('filter', this.deserialize(hashValue, Object));
-					console.log('route changed for filter', name);
+				if (hashValue === value) {
+					return;
 				}
+
+				column.set('filter', this.deserialize(hashValue, Object));
+				console.log('route changed for filter', name);
 			});
+
 		},
 
 		_paramForRouteChanged: function (hashParam, key, value) {
--- a/cosmoz-omnitable.js
+++ b/cosmoz-omnitable.js
@@ -1079,31 +1079,39 @@
 		},
 
 		_paramForRouteChanged: function (hashParam, key, value) {
-			if (hashParam  && this._routeHashParams) {
-
-				var path = ['_routeHashParams', hashParam + '-' + key],
-					hashValue = this.get(path),
-					serialized = this.serialize(value, this.properties[key].type);
-
-				if (serialized !== hashValue) {
-					this.set(path, serialized === undefined ? null : serialized);
-				}
+			if (!hashParam || !this._routeHashParams) {
+				return;
 			}
+
+			let path = ['_routeHashParams', hashParam + '-' + key],
+				hashValue = this.get(path),
+				serialized = this.serialize(value, this.properties[key].type);
+
+			if (serialized === hashValue) {
+				return;
+			}
+
+			this.set(path, serialized === undefined ? null : serialized);
+			console.log('property changed', key, serialized === undefined ? null : serialized);
+
 		},
 
 		_filterForRouteChanged: function (column) {
-			if (this.hashParam && this._routeHashParams && this.data && this.data.length > 0) {
-				var path = ['_routeHashParams', this.hashParam + '-filter--' + column.name],
-					hashValue = this.get(path),
-					value = column.filter,
-					serializedValue = this.serialize(value, Object);
-
-				if (serializedValue !== hashValue) {
-					this.set(path, serializedValue);
-					console.log(column.title, column.filter);
-				}
-
+			if (!this.hashParam || !this._routeHashParams || !this.data || !this.data.length || !this.columns || !this.columns.length)  {
+				return;
 			}
+
+			let path = ['_routeHashParams', this.hashParam + '-filter--' + column.name],
+				hashValue = this.get(path),
+				value = column.filter,
+				serializedValue = this.serialize(value, Object);
+
+			if (serializedValue === hashValue) {
+				return;
+			}
+
+			this.set(path, serializedValue);
+			console.log('filter changed', column.title, column.filter);
 		}
 	});
 }());
--- a/cosmoz-omnitable.js
+++ b/cosmoz-omnitable.js
@@ -200,9 +200,7 @@
 			'_dataChanged(data.*)',
 			'_debounceSortItems(sortOn, descending, filteredGroupedItems)',
 			'_routeHashParamsChanged(_routeHashParams.*, hashParam, columns)',
-			'_paramForRouteChanged( hashParam, "groupOn", groupOn)',
-			'_paramForRouteChanged( hashParam, "sortOn", sortOn)',
-			'_paramForRouteChanged( hashParam, "descending", descending)'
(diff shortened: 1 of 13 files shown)


cosmoz-omnitable.js
```diff
@@ -182,12 +182,25 @@
...
+				type: Object,
+				notify: true
+			},
 		},
 
 		observers: [
 			'_dataChanged(data.*)',
-			'_debounceSortItems(sortOn, descending, filteredGroupedItems)'
+			'_debounceSortItems(sortOn, descending, filteredGroupedItems)',
+			'_routeHashParamsChanged(_routeHashParams.*, hashParam, data)',
```
nomego: These observers observe the same changes, rather chain them if needed?
But I think we should only need one observer for the hash params, that configure affected properties.
Also we shouldn't need to observe data, this feels like a workaround for lifecycle issues?

cosmoz-omnitable.js
```diff
@@ -1008,6 +1021,38 @@
 				return;
 			}
 			gl.highlightItem(i, reverse);
+		},
+
+		_routeHashParamsChanged: function (changes, hashParam, data) {
+			if (! (changes === undefined || hashParam === undefined || data === undefined) && data.length && !this._paramForRouteChanging) {
```
nomego: Rather invert this and return if we miss required values.

cosmoz-omnitable.js
```diff
@@ -1008,6 +1021,38 @@
...
+				this._routeHashParamsInitialized = true;
+				this._routeHashParamsUpdating = true;
+
+				['sortOn', 'groupOn', 'descending'].forEach(function (key) {
```
nomego: Use arrow function to avoid passing thisArg

cosmoz-omnitable.js
```diff
@@ -1008,6 +1021,38 @@
...
+					var path =  ['_routeHashParams', hashParam + '-' + key],
```
nomego: Const

cosmoz-omnitable.js
megheaiulian: Added constant `HASH_PARAMS`.

cosmoz-omnitable.js
megheaiulian: Added constant HASH_PARAMS.

cosmoz-omnitable.js
megheaiulian: Changed this because we actually need `changes|_routeHashParams` and `hashParam` to be truthy values.

cosmoz-omnitable.js
megheaiulian: We don't have to wait for `data` so I removed it. I keep hashParam in observer because when it is the observer will not event fire.

megheaiulian:
@nomego Removed changes to `OmnitableColumnBehavior`.

cosmoz-omnitable.js
```diff
@@ -182,12 +184,26 @@
...
+			},
 		},
 
 		observers: [
 			'_dataChanged(data.*)',
-			'_debounceSortItems(sortOn, descending, filteredGroupedItems)'
+			'_debounceSortItems(sortOn, descending, filteredGroupedItems)',
+			'_routeHashParamsChanged(_routeHashParams.*, hashParam)',
+			'_routeHashParamsForFiltersChanged(_routeHashParams.*, hashParam, columns)',
+			'_paramForRouteChanged( hashParam, "groupOn", groupOn)',
```
nomego: Instead of adding observers here, I think `_paramForRouteChanged` should be called from `_groupItems()` and `_sortFilteredGroupedItems()` respectively. (Probably renamed to something like `updateRouteHashParam()` or similar.
There is a debounce and a delay between `groupOnColumn` being changed and grouping actually happening, which normally means that `groupOnColumn` can change multiple times in a short time and no actual grouping should happen until after a cool-down. If we update routeHashParams during that time, we will be out of sync.
We should probably refactor these debounces away but until then let's keep the routeHashParam-updates close to where the actual action is happening.
It will maintain a simpler, sequential flow with less risk of (introduced) race conditions.

cosmoz-omnitable.js
```diff
@@ -455,6 +471,11 @@
 			return column;
 		},
 
+		_filterChanged: function (e, detail) {
+			this._debounceFilterItems();
+			this._filterForRouteChanged(detail.column);
```
nomego: As with `_paramForRouteChanged()`, this should be called from `_filterItems()` rather, especially since `_filterForRouteChanged()` performs similar checks. I think `updateRouteHashParam()` (or similar) could be adjusted somewhat to also handle filter values.

cosmoz-omnitable.js
```diff
@@ -182,12 +184,26 @@
...
+				notify: true
+			},
 		},
 
 		observers: [
 			'_dataChanged(data.*)',
-			'_debounceSortItems(sortOn, descending, filteredGroupedItems)'
+			'_debounceSortItems(sortOn, descending, filteredGroupedItems)',
+			'_routeHashParamsChanged(_routeHashParams.*, hashParam)',
+			'_routeHashParamsForFiltersChanged(_routeHashParams.*, hashParam, columns)',
```
nomego: Avoid having multiple observers for the same things. `_routeHashParamsForFiltersChanged()` could be called as a helper method when we detect filter changes. `columns` could be observed in for '_routeHashParamsChanged`, if needed.

cosmoz-omnitable.js
```diff
@@ -1008,6 +1029,78 @@
...
+					serialized = this.serialize(value, this.properties[key].type);
+
+				if (serialized !== hashValue) {
+					this.set(path, serialized === undefined ? null : serialized);
+				}
+			}
+		},
+
+		_filterForRouteChanged: function (column) {
+			if (this.hashParam && this._routeHashParams && this.data && this.data.length > 0) {
```
nomego: Invert and return

megheaiulian: Fixed.

cosmoz-omnitable.js
```diff
@@ -1008,6 +1029,78 @@
...
+
+				if (hashValue !== value) {
+					column.set('filter', this.deserialize(hashValue, Object));
+					console.log('route changed for filter', name);
+				}
+			});
+		},
+
+		_paramForRouteChanged: function (hashParam, key, value) {
+			if (hashParam  && this._routeHashParams) {
```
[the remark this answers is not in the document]
megheaiulian: Fixed.

cosmoz-omnitable.js
```diff
@@ -1008,6 +1029,78 @@
...
+					column,
+					value;
+
+				if (!name) {
+					return;
+				}
+				column = this.columns.find(c => c.name === name);
+
+				if (!column) {
+					return;
```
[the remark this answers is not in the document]
megheaiulian: Fixed.

cosmoz-omnitable.js
```diff
@@ -1008,6 +1029,78 @@
...
+					this.set(key, deserialized);
+				}
+			});
+		},
+
+		_routeHashParamsForFiltersChanged: function (changes, hashParam, columns) {
+			if (!changes || !hashParam || !(columns && columns.length)) {
+				return;
+			}
+			Object.keys(changes.base).map(key => {
```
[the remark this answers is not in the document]
megheaiulian: Fixed.

cosmoz-omnitable.js
```diff
@@ -1008,6 +1029,78 @@
...
+					deserialized = this.deserialize(hashValue, this.properties[key].type);
+
+				if (hashValue !== undefined && deserialized !== this.get(key)) {
+					this.set(key, deserialized);
+				}
+			});
+		},
+
+		_routeHashParamsForFiltersChanged: function (changes, hashParam, columns) {
+			if (!changes || !hashParam || !(columns && columns.length)) {
```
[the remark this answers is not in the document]
megheaiulian: Fixed.

cosmoz-omnitable.js
megheaiulian: Changed to use only _routeHashParamsChanged.

cosmoz-omnitable.js
megheaiulian: In `_filterForRouteChanged()` I handle the change only on the column that changed filter.
In filterItems there is no info about the changed column.

nomego: True.
We should probably track individual column filter changes for the actual filtering as well at some point.

cosmoz-omnitable.js
```diff
@@ -1008,6 +1026,93 @@
...
+
+				if (hashValue === undefined ||  deserialized === this.get(key)) {
+					return;
+				}
+
+				this.set(key, deserialized);
+				console.log('route changed for', key, deserialized);
+			});
+
+			let rule = new RegExp('^' + hashParam + '-filter\-\-([a-z0-9\-]+)$'),
```
nomego: Const

cosmoz-omnitable.js
```diff
@@ -1008,6 +1026,93 @@
...
+				const matches = key.match(rule),
+					name = matches && matches[1],
+					hashValue = routeParams[key];
+				let column,
+					value;
+
+				if (!name) {
+					return;
+				}
+				column = this.columns.find(c => c.name === name);
```
nomego: I wouldn't mind moving the let declarations here, what do you think?

cosmoz-omnitable.js
```diff
@@ -182,12 +184,23 @@
...
+				notify: true
+			},
 		},
 
 		observers: [
 			'_dataChanged(data.*)',
-			'_debounceSortItems(sortOn, descending, filteredGroupedItems)'
+			'_debounceSortItems(sortOn, descending, filteredGroupedItems)',
+			'_routeHashParamsChanged(_routeHashParams.*, hashParam, columns)',
+			'_paramsForRouteChanged( groupOn, sortOn, descending)'
```
nomego: Won't this introduce the issue you mentioned with observer not being triggered if any of the values are undefined?

cosmoz-omnitable.js
```diff
@@ -1008,6 +1031,102 @@
...
+		},
+
+		_serializeFilter: function (obj) {
+			const type = {}.toString.call(obj).split(' ')[1].slice(0, -1).toLowerCase();
+			let value = obj;
+
+			if (type === 'array' && !value.length) {
+				value =  null;
+			} else if (type === 'object') {
+				value =  Object.keys(obj).filter(k => obj[k] != null)
```
nomego: I would rewrite like this for readability, what do you think?
Having a default value for acc and also providing an initial value seems contradictory.
It should also never evaluate to falsy?
Or is this a way to make it undefined if there are no keys?
I would want that a lot clearer then.

```js
const setKeys = Object.keys(obj).filter(k => obj[k] != null);
if (setKeys.length === 0) {
  value = null;
} else {
  value = setKeys.reduce((acc, k) => {
    acc[k] = obj[k];
    return acc;
  }, {});
}
```

megheaiulian: @nomego Fixed.

megheaiulian:
@nomego Added basic hash-param tests.
 PR ready for final review.

Issue #69, opened 2017-10-24
filter property of date column is readonly


plequang:
In comoz-omnitable-date-behavior, the `filter` property is redefined as :
```js
filter: {
	type: Object,
	computed: '_computeFilter(_filterInput.start, _filterInput.end)'
},
```
this makes the `filter` property readonly, preventing users to set the filter from outside by binding to this property.
One should be able to bind to the filter property of date column the same way as other column types.

nomego:
Agreed, although we should hold off this work until #59 is solved.

PR #73, opened 2017-10-27
Changing groupOn should not overwrite visibleColumns

### LINKED ISSUE #71: Changing groupOn should not overwrite visibleColumns
When `groupOn` property is changed the compute function `_computeVisibleColumns` sets `visibleColumns` to a new Array.
This triggers the removal and addition of a lot of elements when only a column has been removed.
 `_computeVisibleColumns` should be changed to a observer that does micro updates to `visibleColumns` (add/removes one column from the Array).

megheaiulian:
fixes #71 
The groupOnColumn should be excluded from filters.

### DIFF
--- a/cosmoz-omnitable-column-behavior.html
+++ b/cosmoz-omnitable-column-behavior.html
@@ -353,6 +353,7 @@
 			}
 			if (typeof reset !== 'function') {
 				this.filter = reset;
+				return;
 			}
 			this.filter = reset();
 		},
--- a/cosmoz-omnitable-column-behavior.html
+++ b/cosmoz-omnitable-column-behavior.html
@@ -347,6 +347,8 @@
 		resetFilter: function () {
 			const reset = this.get('properties.filter.value');
 			if (reset === undefined) {
+				// Until date| columns are fixed we must trigger a notification for grouping to happen
+				this.__filterChanged();
 				return;
 			}
 			if (typeof reset !== 'function') {
--- a/cosmoz-omnitable-column-behavior.html
+++ b/cosmoz-omnitable-column-behavior.html
@@ -344,6 +344,17 @@
 			return this._applyMultiFilter.bind(this, this.filter);
 		},
 
+		resetFilter: function () {
+			const reset = this.get('properties.filter.value');
+			if (reset === undefined) {
+				return;
+			}
+			if (typeof reset !== 'function') {
+				this.filter = reset;
+			}
+			this.filter = reset();
+		},
+
 		_applySingleFilter: function (filterString, item) {
 			var value = this.get(this.valuePath, item).toString().toLowerCase();
 			return value.indexOf(filterString) >= 0;
--- a/cosmoz-omnitable-repeater-behavior.html
+++ b/cosmoz-omnitable-repeater-behavior.html
@@ -16,6 +16,10 @@
 			columns: {
 				type: Array
 			},
+			groupOnColumn: {
+				type: Object,
+				observer: '_groupOnColumnChanged'
+			}
 		},
 
 		observers: [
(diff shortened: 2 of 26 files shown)


cosmoz-omnitable.js
```diff
@@ -256,13 +256,21 @@
 
 		_scalingUp: false,
 
-		_computeVisibleColumns(columns, groupOn) {
-			if (!columns) {
-				return;
+		_groupOnColumnChanged: function (column, previousColumn) {
+			let index;
+
+			if (previousColumn && (index = this.columns.indexOf(previousColumn)) > -1)  {
```
nomego: For readability, do

```js
if (previoiusColumn) {
  let index = ..
}
if (column) {
  let index = ..
}
```

it will be scoped within the if anyway

nomego: I don't think we can guarantee that `visibleColumns` and `columns` can cross-use indexes?

megheaiulian: @nomego All this will be changed as discussed. We will not actually remove groupOnColumn from visibleColumns but we'll instead forward it to all repeaters and they simply hide it.

megheaiulian: Same as above.

megheaiulian:
Tests are checking for changes to `visibleColumns` when `groupOn` is changed. This no longer needs to happen.

cosmoz-omnitable-column-behavior.html
```diff
@@ -344,6 +344,19 @@
...
 
+		resetFilter: function () {
+			const reset = this.get('properties.filter.value');
+			if (reset === undefined) {
+				// Until date| columns are fixed we must trigger a notification for grouping to happen
+				this.__filterChanged();
+				return;
+			}
+			if (typeof reset !== 'function') {
+				this.filter = reset;
```
nomego: We should probably return here?

cosmoz-omnitable-repeater-behavior.html
```diff
@@ -197,6 +201,24 @@
 			this._elements.splice(start, removedColumns.length);
 		},
 
+		_groupOnColumnChanged: function (column, previousColumn) {
+			if (!this._elements || !this._elements.length) {
+				return;
+			}
+			if (previousColumn) {
+				let previousElement = this._elements.find(e => e.__column === previousColumn);
```
nomego: const, same below

cosmoz-omnitable-column-behavior.html
megheaiulian: Return added.

megheaiulian:
@nomego Can you have a look again?

cosmoz-omnitable-column-behavior.html
plequang: So just to be sure, now the filter is reset for the groupOn column, as discussed in #74.
Can you confirm @nomego ?

(Except it does not work for date column because of #69)

If going this path, wouldn't it be nicer to have a `_getDefaultFilter()` function in each column, and have the default filter value look like:
```
filter: {
   value: function () {
      return this._getDefaultFilter();
   }
}
```
then in `resetFilter`
```
this.filter = this._getDefaultFilter();
```

cosmoz-omnitable-repeater-behavior.html
```diff
@@ -197,6 +201,24 @@
...
+				if (previousElement) {
+					previousElement.removeAttribute('hidden');
+				}
+			}
+			if (column) {
+				let element = this._elements.find(e => e.__column === column);
+				if (element) {
+					element.setAttribute('hidden', '');
```
plequang: Maybe we should secure `display: none` for slotted element that have the `hidden` attribute in all repeaters.

cosmoz-omnitable-column-behavior.html
nomego: Yes we can start like that and do the bigger solution down the line if we want to change the behavior.

`_getDefaultFilter()` could be a nice option too. I would prefer JS setters/getters but it wasn't great with Polymer 1.x apparently.

PR #80, opened 2017-10-31
Add _getDefaultFilter and CSS for hidden attribute in repeaters

### LINKED ISSUE #69: filter property of date column is readonly

megheaiulian:
Added _getDefaultFilter and CSS for hidden attribute in repeaters
Added _getDefaultFilter for date columns
Filter for date columns is not computed and readonly anymore
Check for null filter.start & filter.end in date columns getFilterFn

fixes #69

### DIFF
--- a/cosmoz-omnitable-column-behavior.html
+++ b/cosmoz-omnitable-column-behavior.html
@@ -46,7 +46,10 @@
 
 			filter: {
 				type: Object,
-				notify: true
+				notify: true,
+				value: function () {
+					return this._getDefaultFilter();
+				}
 			},
 
 			/**
@@ -345,17 +348,24 @@
 		},
 
 		resetFilter: function () {
-			const reset = this.get('properties.filter.value');
-			if (reset === undefined) {
-				// Until date| columns are fixed we must trigger a notification for grouping to happen
-				this.__filterChanged();
-				return;
+			this.filter = this._getDefaultFilter();
+		},
+
+		_getDefaultFilter: function () {
+			const type = this.get('properties.filter.type'),
+				value = this.get('filter');
+
+			if (type === Object && value != null && value.constructor === String) {
+				return null;
 			}
-			if (typeof reset !== 'function') {
-				this.filter = reset;
-				return;
+
+			if (type === Array) {
+				return [];
 			}
-			this.filter = reset();
+
+			if (type === Boolean) {
+				return null;
+			}
 		},
 
 		_applySingleFilter: function (filterString, item) {
--- a/cosmoz-omnitable-column-behavior.html
+++ b/cosmoz-omnitable-column-behavior.html
@@ -47,7 +47,7 @@
 			filter: {
 				type: Object,
 				notify: true,
-				value: function () {
+				value() {
 					return this._getDefaultFilter();
 				}
 			},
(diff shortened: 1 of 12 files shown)


cosmoz-omnitable-column-behavior.html
```diff
@@ -345,17 +348,24 @@
...
-				this.__filterChanged();
-				return;
+			this.filter = this._getDefaultFilter();
+		},
+
+		_getDefaultFilter: function () {
+			const type = this.get('properties.filter.type'),
+				value = this.get('filter');
+
+			if (type === Object && value != null && value.constructor === String) {
```
nomego: Why do we specifically need to condition string filter for object ?

nomego:
Also, something this doesn't consider is if we're setting `filter` for a date column, it filter inputs will not be updated.

cosmoz-omnitable-column-behavior.html
megheaiulian: It is needed for basic columns to update the text input. When set to null it doesn't get set to null.

megheaiulian: @nomego I've refactored this.

cosmoz-omnitable-column-date-behavior.html
```diff
@@ -246,18 +263,23 @@
...
 				return null;
 			}
-			if (!(value instanceof Date)) {
-				return new Date(value);
+			let date = value;
+			if (!(date instanceof Date)) {
+				date = new Date(date);
 			}
-			return value;
+			if (isNaN(date)) {
```
nomego: The idea was to use `_isValidDate()` for these kind of checks. Any reason not to?

cosmoz-omnitable-column-datetime.html
```diff
@@ -14,7 +14,7 @@
 
 		<template class="edit-cell">
 			<datetime-local no-label-float type="datetime-local" on-change="_dateValueChanged"
-				value="[[ renderISODateTimeValue(item, valuePath) ]]">
+				value="[[ renderISODateValue(item, valuePath) ]]">
```
nomego: This method had "time" in its name to highlight that it does not only produce a date.
I see that you re-use the inherited method but then it needs to be even more abstract to be semantically right. `renderISOValue` maybe? But I think the `DateTime` part adds a lot of clarity/readability.

nomego:
@plequang What rendering issues? 'Invalid Date'? Just noticed that one, will fix now.

cosmoz-omnitable-column-datetime.html
megheaiulian: Both `renderISODateValue` and old `renderISODateTimeValue` were calling very similar functions. Moved that into `formatISODate`.  This also highlights that there is no difference between time and date except the way we render it.

cosmoz-omnitable-column-date-behavior.html
megheaiulian: Either `getDate` or `isValidDate` should go away as they are too similar. Also calling `getDate` with a String or another invalid value you would get that value back instead of null.

cosmoz-omnitable-column-time.html
```diff
@@ -144,14 +144,30 @@ <h3 style="margin: 0;">[[ title ]]</h3>
...
+				}
+
+				this._updatingInput = true;
+				this._filterInput = { start, end};
+				this._updatingInput = false;
+			},
+
+			_filterInputChanged: function (start, end) {
+				const filter = this.filter;
+				if (!filter || this._updatingInput || start === filter.start && end  === filter.end) {
```
nomego: We shouldn't need _updatingInput since start should be filter.start and end should be filter.end?

PR #81, opened 2017-10-31
Fix toXlsxValue in autocomplete column

### LINKED ISSUE #79: Objects in XLSX
When exporting an object to XLSX, the cell data becomes `[object Object]`, but will be exported properly in CSV.

Look at full demo, export a few lines to XLSX, and check the "Object" column.

megheaiulian:
Fix toXlsxValue in autocomplete column
toXlsxValue uses textProperty
fixes #79

### DIFF
--- a/cosmoz-omnitable-column-autocomplete.html
+++ b/cosmoz-omnitable-column-autocomplete.html
@@ -84,9 +84,11 @@
 					return;
 				}
 				const itemValue = this.get(valuePath, item),
-					type = {}.toString.call(itemValue).split(' ')[1].slice(0, -1).toLowerCase(),
 					textProp = this.textProperty;
-				if (textProp && type === 'object') {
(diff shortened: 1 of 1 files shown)


cosmoz-omnitable-column-autocomplete.html
```diff
@@ -84,13 +84,21 @@
 					return;
 				}
 				const itemValue = this.get(valuePath, item),
+					type = {}.toString.call(itemValue).split(' ')[1].slice(0, -1).toLowerCase(),
```
nomego: Why? Seems expensive. Should be in an `_isObject()` helper at least?

megheaiulian: This is the only way to check that it is actually an Object. `typeof []' is `object`.

PR #82, opened 2017-10-31
Sort groups descending

megheaiulian:
Sort groups descending

### DIFF
--- a/cosmoz-omnitable.js
+++ b/cosmoz-omnitable.js
@@ -109,11 +109,6 @@
 				value: false
 			},
 
-			sortDirection: {
-				type: String,
-				computed: '_computeSortDirection(descending, t)'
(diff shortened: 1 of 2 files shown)


cosmoz-omnitable.js
```diff
@@ -967,6 +976,15 @@
 			}
 			this.descending = false;
 		},
+
+		_reverseGroupOnDirection(e) {
+			var column = e.model.column;
+			if (column.name === this.groupOn) {
```
nomego: Rather compare with groupOnColumn?

megheaiulian:
@nomego Need to check that it doesn't trigger more than one  `groupItems` call.

megheaiulian:
@nomego  Ready for review.

PR #86, opened 2017-11-02
Add and remove hidden attribute for groupOnColumn elements

### LINKED ISSUE #85: groupOn column is not hidden for item rows instantiated for the first time
# Description

In the `full-demo` page, when setting a `group-on`  column, this column is not hidden from the header row and item rows.
For example, using 
```
<cosmoz-omnitable ....  group-on="group">
```
Will produce

image

When changing "Group on" to "No grouping" then to "Group" again, the column is correctly hidden.

This cause also problems when changing the group on column: if new `cosmoz-omnitable-item-row` elements are created to render the table, they won't have the group column hidden

megheaiulian:
When creating or removing elements in repeater add and remove the hidden attribute.

fixes #85

### DIFF
--- a/cosmoz-omnitable-repeater-behavior.html
+++ b/cosmoz-omnitable-repeater-behavior.html
@@ -16,6 +16,7 @@
 			columns: {
 				type: Array
 			},
+
 			groupOnColumn: {
 				type: Object,
 				observer: '_groupOnColumnChanged'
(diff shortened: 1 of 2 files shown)


megheaiulian:
@plequang This is ready for review.

cosmoz-omnitable-repeater-behavior.html
```diff
@@ -162,6 +162,13 @@
 				instance = this._getTemplateInstance(column);
 				element.__instance = instance;
 				element.__column = column;
+
+				if (column === this.groupOnColumn) {
+					element.setAttribute('hidden', '');
+				} else if (element.hasAttribute('hidden')) {
+					element.removeAttribute('hidden', '');
+				}
+
```
plequang: But the `hidden` attribute should not be added to the element in `cosmoz-omnitable-group-row` : 

image

megheaiulian: Changed it. No longer passing group-on-column to group-row.

Issue #62, opened 2017-10-13
Use NumberFormat to render amounts


nomego:
As with the number-column, the amount column should use the Intl.NumberFormat to render amounts in user's browser locale.

Related to #24

PR #88, opened 2017-11-14
Amount formatting now uses intl.numberformat. refs #62

JockeCK:


cosmoz-omnitable-column-amount.html
```diff
@@ -67,12 +65,16 @@ <h3 style="margin: 0;">[[ title ]]</h3>
...
-					computed: '_computeFilterText(filter.*)'
+					computed: '_computeFilterText(filter.*, _formatters)'
 				},
 				_possibleAmountRange: {
 					type: Object,
 					computed: '_computePossibleAmountRange(values.*)'
 				},
+				_formatters: {
+					type: Object,
+					computed: '_computePossibleFormatters(locale, values.*)'
```
nomego: Do we really need to observe .*?

cosmoz-omnitable-column-amount.html
```diff
@@ -198,18 +222,14 @@ <h3 style="margin: 0;">[[ title ]]</h3>
 
 			_setMinValue: function (event) {
 				const value = this._getNumberValue(event.target.value);
-				this.set('filter.min', {
```
nomego: Why were these changes needed ?

cosmoz-omnitable-column-amount.html
```diff
@@ -219,19 +239,19 @@ <h3 style="margin: 0;">[[ title ]]</h3>
 				return accounting.formatNumber(item.amount, item.currency);
 			},
 
-			getString: function (item, valuePath) {
+			getString: function (item, valuePath, _formatters = this._formatters) {
 				var value = this.get(valuePath || this.valuePath, item);
-				if (value) {
-					return this.renderAmount(value);
+				if (value && _formatters && _formatters[value.currency]) {
```
nomego: This logic is inverted compared to toXlsxValue, and since it's needed in mulitple places I think it should be moved to a helper method.
Actually, `renderAmount` (or `formatAmount`?) should probably be implemented locally `(value, _formatters = this.formatters)` to be used wherever we use formatters to format an amount.

PR #89, opened 2017-11-14
Don't handle hash param updates while updating route params fix #87

### LINKED ISSUE #87: Group on filtered value
When filter in header in column and then group on the same column header. The value in the column header disappear and the omnitable group on another column header.

megheaiulian:
related to #87 
When `groupOnColumn` changes in the same call stack `resetFilter` is called and hash params are updated. In the same stack than route hash params are checked and groupOn is different so it gets set to the value from hash-param.

fixes #87

### DIFF
--- a/cosmoz-omnitable.js
+++ b/cosmoz-omnitable.js
@@ -520,6 +520,7 @@
 		},
 
 		_groupOnColumnChanged: function (column) {
+			this._updateRouteParam('groupOn');
 			if (column && column.filter) {
 				column.resetFilter();
 			} else {
@@ -541,8 +542,6 @@
 				this._groupsCount = 0;
 				return;
 			}
-
-			this._updateRouteParam('groupOn');
 
 			var groupOnColumn = this.groupOnColumn,
 				groups;


cosmoz-omnitable.js
```diff
@@ -1196,8 +1195,13 @@
...
 				return;
 			}
+			this._updateRoute(path, serialized);
+		},
 
-			this.set(path, serialized === undefined ? null : serialized);
+		_updateRoute(path, value) {
+			this._updatingRoute = true;
+			this.set(path, value === undefined ? null : value);
+			this._updatingRoute = true;
```
nomego: Does this ever get set to false?

cosmoz-omnitable.js
```diff
@@ -1196,8 +1195,13 @@
 			if (serialized === hashValue) {
 				return;
 			}
+			this._updateRoute(path, serialized);
+		},
 
-			this.set(path, serialized === undefined ? null : serialized);
+		_updateRoute(path, value) {
```
nomego: Default value of value could be null to make set cleaner.

megheaiulian:
@nomego Updated the PR. 
First commit fixes the direct issue.
The second commit contains the _updateRoute method where we set the flag to suppress notifications in `_routeHashParamsChanged`.

nomego:
I will merge this, but personally I think these control vars is a way to suppress behavior, and if we need to do that, the behavior is wrong.
I would rather see that we put our efforts into `_routeHashParamsChanged` to make it as performant as possible, so nothing is triggered and performance impact is minimal if it's get called without any URL params / omnitable properties being out of sync.
I think one change in this direction is to not act on `changes.base` every time but rather `changes.path`. This would avoid the whole loop at the end and give more performance than adding a control var/flag.

megheaiulian:
@nomego I agree that `changes.path` would be a lot better. The problem is `iron-location` and `cosmoz-page-location` don't handle micro update and instead the full object is changed so we have at the moment no way to check `changes.path`.

megheaiulian:
@nomego I think at the moment I rebase this to include only the group fix. `cosmoz-page-location` should in turn not fire a update or fire it diferently when routeHashParams are changed from outside. Basically to handle and fix this in `cosmoz-page-router`.
Let me know what you think about that.

Issue #78, opened 2017-10-28
Dialog for hidden filters


nomego:
When columns get disabled, their filter possibilities disappear.
I propose a filter icon in the header row above the "expand row" area, see screenshot.
When clicked, a cosmoz-dialog will show all the missing filters for the user to configure.

![image](https://user-images.githubusercontent.com/960553/32138623-1d33d922-bc36-11e7-8a17-cb46206be875.png)

Hopefully approved by  @Neovici/uiux

PR #90, opened 2017-11-17
Dialog for hidden filters [issue-78]

JaySunSyn:
See https://github.com/Neovici/cosmoz-omnitable/issues/78

Please merge https://github.com/Neovici/cosmoz-omnitable/pull/117 first


cosmoz-omnitable.html
```diff
@@ -52,7 +54,15 @@
 					<paper-checkbox checked$="{{ _allSelected }}" on-change="_onAllCheckboxChange" hidden$="[[ !_dataIsValid ]]">
 					</paper-checkbox>
 				</div>
-				<cosmoz-omnitable-header-row columns="[[ visibleColumns ]]" group-on-column="[[ groupOnColumn ]]"></cosmoz-omnitable-header-row>
+				<cosmoz-omnitable-header-row columns="[[ visibleColumns ]]" group-on-column="[[ groupOnColumn ]]">
+					<div slot="suffix" class="align-right" hidden$="[[allVisible]]">
```
plequang: This breaks `_adjustHeadersWidth`.
In `_adjustHeadersWidth`, we reference individual header cell by their index in header row children (see https://github.com/Neovici/cosmoz-omnitable/blob/master/cosmoz-omnitable.js#L786)
But now, the `suffix` div is the first children in header row, so we don't size the right element anymore

cosmoz-omnitable-filter-dialog.html
```diff
@@ -0,0 +1,58 @@
...
+            <div id="filters"></div>
+    </template>
+    <script>
+    (() => {
+    	'use strict';
+    	Polymer({
+    		is: 'cosmoz-omnitable-filter-dialog',
+    
+    		behaviors: [
+    			Polymer.PaperDialogBehavior,
```
plequang: We can't use paper-dialog, because it has stacking context issues (the dialog can appear below another element).
`cosmoz-dialog` solves stacking context issues, but works differently.
Instead, you should use 
```
<cosmoz-dialog>
   <template>
      dialog content
   <template>
 ```

Regarding the dialog content, you should investigate using `cosmoz-omnitable-repeater-behavior`, as this behavior already handles columns changes and instances creation/release. 
See for example https://github.com/Neovici/cosmoz-omnitable/blob/master/cosmoz-omnitable-header-row.html

File: cosmoz-omnitable.html [hunk 3]
```diff
@@ -195,7 +205,13 @@ <h3 class="groupRow-label">
 		<div id="columns">
 			<slot></slot>
 		</div>
-
+		
+		<cosmoz-omnitable-filter-dialog
```
plequang: So this should rather looks like (relating to other comments):
```
<cosmoz-dialog with-backdrop>
   <template>
       <cosmoz-omnitable-filters-dialog columns="[[disabledColumns]]">
   <template>
 ```

File: cosmoz-omnitable.html (hunk 1 above)
Comment (JaySunSyn) [reply cosmoz-omnitable.html]: Thanks, pls let me know if the changes work for you.

File: cosmoz-omnitable.js [hunk 4]
```diff
@@ -297,6 +304,14 @@
 			return dataIsValid && hasActions;
 		},
 
+		_computeColumnsAllVisible(visibleColumnsChange, columns) {
```
plequang: This does not work for the groupOn column.
When grouping on a column, it is not removed from the `visibleColumn` property, but the cells for the groupOn colunm are hidden.
The `_computeColumnsAllVisible` function could  also observe changes to the `groupOnColumn` property, and If this is not null, then it should return false

cosmoz-omnitable.html
```diff
@@ -196,6 +206,16 @@ <h3 class="groupRow-label">
 			<slot></slot>
 		</div>
 
+		<cosmoz-dialog with-backdrop id="filterDialog">
```
plequang: I think the dialog needs at least a "Close" button. 
It's not always clear to the user that clicking somewhere outside the dialog will close it.

cosmoz-omnitable-headers-dialog.html
```diff
@@ -0,0 +1,81 @@
...
+
+			_columnsChanged(columnsChange) {
+				const columns = columnsChange.base,
+					headers = this.$.headers;
+
+				if (!columns) {
+					return;
+				}
+
+				headers.innerHTML = '';
```
plequang: This is not the right way to clear the content of the dialog.
`cosmoz-omnitable-templatizer` keeps a reference to the instances it creates, so when an instance is not used anymore it needs to be "released", otherwise it will stay in memory.
This is also not very efficient, as we recreate new instances for all headers in the dialog when `disabledColumns`  is changing. 

You should really try to use `cosmoz-omnitable-repeater-behavior`, as it takes care of correctly releasing unused instances as well as creating only instances when needed.
`cosmoz-omnitable-headers-dialog` could be implemented the same way as `cosmoz-omnitable-header-row`.

JaySunSyn:
Please check again.

cosmoz-omnitable-headers-dialog.html
```diff
@@ -0,0 +1,99 @@
...
+In the above example, the `<cosmoz-omnitable-headers-dialog>` element is wrapped
+with a `cosmoz-dialog` element which enables it to be displayed as a dialog.
+
+@group Cosmoz Elements
+@element cosmoz-omnitable-headers-dialog
+@demo demo/index.html
+-->
+<dom-module id="cosmoz-omnitable-headers-dialog">
+	<template>
+		<style>
```
plequang: Styling of the dialog looks strange.
host might require a display block.
container should use a vertical flexbox.

cosmoz-omnitable-headers-dialog.html
```diff
@@ -0,0 +1,99 @@
...
+				font-size: 1.5em;
+    			font-weight: 600;
+				@apply --cosmoz-omnitable-headers-dialog-title;
+			}
+		</style>
+
+		<paper-icon-button class="close-btn" icon="close" on-tap="_close"></paper-icon-button>
+		
+		<div class="container">
+			<div class="title">[[title]]</div>
```
plequang: Remove the title from here, and use `<h2>...</h2>` in the dialog template in omnitable.html
That way, it will use paper-dialog default styling.

cosmoz-omnitable.html
```diff
@@ -196,6 +208,17 @@ <h3 class="groupRow-label">
...
+			<template>
```
plequang:
```
<h2>[[_('Filter', t)]]</h2>
```

cosmoz-omnitable.html
```diff
@@ -196,6 +208,17 @@ <h3 class="groupRow-label">
...
+				<cosmoz-omnitable-headers-dialog
+					title="[[_('Filter', t)]]"
+					columns="[[_filterDialogColumns]]"
+					on-close-tapped="_closeFilterDialog"
+					with-backdrop>
+				</cosmoz-omnitable-headers-dialog>
```
plequang:
```
<div class="buttons>
<paper-button>[[_("Close")]]</paper-button>
</div>
```

cosmoz-omnitable-headers-dialog.html
```diff
@@ -0,0 +1,99 @@
...
+			}
+
+			.title {
+				font-size: 1.5em;
+    			font-weight: 600;
+				@apply --cosmoz-omnitable-headers-dialog-title;
+			}
+		</style>
+
+		<paper-icon-button class="close-btn" icon="close" on-tap="_close"></paper-icon-button>
```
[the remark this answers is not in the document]
JaySunSyn: Hmm, took the facebook approach here.. But I thinks you are right..

cosmoz-omnitable-headers-dialog.html
```diff
@@ -0,0 +1,73 @@
...
+
+			_detachTemplateInstance(instance, column, element) {
+				column.headerTemplatizer.detachInstance(instance, element);
+			},
+
+			_configureTemplateInstance(instance) {
+				this._forwardPropertiesFlush(instance);
+			},
+
+			_close() {
```
plequang: Since the close button is now in omnitable.html, is this function still needed ?

plequang:
Some screenshot to illustrate the UI/UX issues:

Single filter with paper-autocomplete-chips:

image

Icon overlaying header:

image

Even more overlaying icon, when there is no scrollbar in the omnitable

image

JaySunSyn:
The not-overlapping suggestions of autocomplete-chips is fixed with this PR https://github.com/Neovici/paper-autocomplete-chips/pull/6

cosmoz-omnitable.html
```diff
@@ -196,6 +209,21 @@ <h3 class="groupRow-label">
 			<slot></slot>
 		</div>
 
+		<cosmoz-dialog with-backdrop id="filterDialog">
+			<template>
+				<h2>[[_('Filter', t)]]</h2>
+				<paper-dialog-scrollable>
```
plequang: If you want to use `paper-dialog-scrollable`, you have to let it manage scroll, meaning `cosmoz-omnitable-headers-dialog` should overflow.

JaySunSyn:
No worries.
What do you think regarding the sorting of the columns in the dialog. I have now doubts if we should sort on the node type but rather just display them as they come in..

cosmoz-omnitable.js
```diff
@@ -610,8 +598,12 @@
...
 
-		_groupOnColumnChanged: function () {
-			this._updateRouteParam('groupOn');
+		_groupOnColumnChanged: function (column) {
+			if (column && column.hasFilter()) {
+				column.resetFilter();
+			} else {
+				this.debounce('groupItems', this._groupItems);
+			}
 			this.debounce('groupItems', this._groupItems);
```
plequang: This is is not necessary (`resetFilter` should in the end trigger a call to groupItems)

cosmoz-omnitable.js
```diff
@@ -308,9 +307,6 @@
 		_sortFilterDialogColumns(a, b) {
```
plequang: As you suggested, let's drop the sorting of filter based on the type.

plequang:
Regarding https://github.com/Neovici/paper-autocomplete-chips/pull/6, could we find a solution that would work without this PR ?

Maybe in `cosmoz-omnitable-headers-dialog`, we could override `_configureElement` with something like
```
_configureElement: function (element, column) {
	if (column.is === 'cosmoz-omnitable-column-autocomplete') {
		do something with paper-autcomplete-chips element so that 
		the suggestion will overlay the dialog.
		What about setting position:fixed on the element ?           
	}
}
```

Not sure if it's possible though.

JaySunSyn:
I've made non-breaking changes to https://github.com/Neovici/paper-autocomplete-chips/pull/6.
- exposed css mixins to style suggestion wrapper

I use those mixins in this PR to update the suggion-wrapper style. This is the only way so far I made it work for both Polymer 1 & 2.

Please let me know what you think.
Thanks

cosmoz-omnitable-headers-dialog.html
```diff
@@ -0,0 +1,95 @@
...
+			_onNodesChange() {
+				const accs = Polymer.dom(this).querySelectorAll('paper-autocomplete-chips');
+	
+				setTimeout(() => {
+					accs.forEach(acc => {
+						const width = `${acc.offsetWidth}px`,
+							position = 'fixed';
+	
+						if (Polymer.Element) {
+							acc.updateStyles({
```
plequang: This can be used in Polymer 1.x as well
https://www.polymer-project.org/1.0/docs/api/Polymer.Base#method-updateStyles

JaySunSyn: Interesting. Thanks.

JaySunSyn:
Does it look like this on your screen or different?

image

plequang:
With Polymer 1.x, this looks like that:

image

It works with Polymer 2.x (but omnitable has a problem in 2.x because `datetime-local` does not handle undefined values in `_setMinMax`).

JaySunSyn:
Hmm :/ I cannot reproduce.. May you check again if you have integrated the PR https://github.com/Neovici/paper-autocomplete-chips/pull/6 and pull also this PR again please. Thanks

plequang:
OK my bad. I was using bower link paper-autocomplete-chips, which links to `bower_components` directory, and not the `bower_components-1.x` directory.

plequang:
Found an issue. To reproduce, follow these steps in the full demo:
1. Reduce the width of the screen to let the filter icon appear
2. open the filter dialog, the close it
3. reduce more the width, so that new filters are added to the dialog.
4. open the dialog

=> the paper-autocomplete will have a width of 0
When the dialog is opened for the first time, `_onNodesChange` is called while the dialog is visible, so `offsetWidth` returns something different than 0
When you reduce the width at step 3, the dialog is not visible, so in `_onNodesChange`, `offsetWidth` will return 0.

plequang:
Found another issue, not related to paper-autocomplete:

1. reduce the width of the screen to let the filter icon appear
2. open the filter dialog
3. maximize the window

=> the dialog is displayed without any filter

cosmoz-omnitable-headers-dialog.html
```diff
@@ -0,0 +1,90 @@
...
+			_elementType: 'div',
+
+			_slotName: 'header-cell',
+
+			listeners: {
+				'iron-resize': '_onResize'
+			},
+
+			_onResize() {
+				this.scopeSubtree(this.$.container);
```
plequang: What's the purpose of this ?

Actually, this seems to break styling of the suggestion wrapper, which does not overlap anymore (but it works if I remove the `scopeSubtree` call)

JaySunSyn: Hmm :/ yes, I see this too now..
It's to apply the styles of `.header-clear-filter` to some intputs

Without:
image

But it seems that we should not use `scopeSubtree` for polymer elements. Only for external libs..

Looking into it.

JaySunSyn:
Added a clear button in PR https://github.com/Neovici/cosmoz-omnitable/pull/117 to apply styles correctly.

JaySunSyn:
Please merge https://github.com/Neovici/cosmoz-omnitable/pull/135 first so building succeeds again.

Issue #60, opened 2017-10-12
Column filter change method configuration properties


nomego:
There are at least four different approaches to indicating filter changes, and currently we mix them somewhat:

## Instant
As soon as a filter changes in any way, the `filter` setting is updated and any bindings/observers are updated/invoked.
This is the current approach for autocomplete, boolean, date, datetime, list, time and also a "regular" column (string search)

## on-change / blur
Updates the filter when one of the filter inputs loses focus or activates a change event (enter?).
Currently used for amount

## Delayed with debounce
Like 'instant', but there is a (renewed) timeout after each change to allow for multiple changes at once.
Currently used for number (since using the +/- buttons for a number input triggers 'on-change')

## Complete change
For filters that open a menu and have multiple inputs (amount, number, date, etc) it might be natural and efficient to not trigger change until the menu closes to allow for all filter-inputs to be updated.
It would make sense to use when the filter will be used to query a backend for new data lists.
Not used right now.

Do we need these methods to be configurable for each column or should we set different methods for different types of column filters and stick to them?

nomego:
# Strategies

## Range (Amount, Number columns)

  - From: Value/query should update on blur
  - From: Enter, tab will move to "To" field
  - To: Value/query should update on blur, menu should close
  - To: Enter will validate input in max/min range
    - If outside range, update with valid value, perform query
    - If inside range, (blur to perform query), close menu
  - From/To: Add clear icon
  - Menu - Focus in "From" when opened

Possibly later: Long debounce

## Plain column (text)

  - Blur, enter to update value
  - Long debounce (1000ms?) to update value

Possibly later: Escape to clear input

## Autocomplete

  - Preselect first suggestion
  - Enter will display suggestions if not displayed (to combat closing suggestions with esc)

## Boolean

TDB

## Date

TDB

## Datetime

TDB

## Time

TDB

## List

TDB

## List horizontal

TDB

PR #94, opened 2017-11-23
Refactor amount and number columns

megheaiulian:
- [x] Synchronize filter with hash
- [x] Refactor duplicate code between amount and number columns

### DIFF
--- a/cosmoz-omnitable-column-range-behavior.html
+++ b/cosmoz-omnitable-column-range-behavior.html
@@ -44,6 +44,11 @@
 				computed: '_computeLimit(_range, _filterInput.*, min, max)'
 			},
 
+			locale: {
+				type: String,
+				value: null
+			},
+
 			_tooltip: {
 				type: String,
 				computed: '_computeTooltip(title, _filterText)'
--- a/cosmoz-omnitable-column-range-behavior.html
+++ b/cosmoz-omnitable-column-range-behavior.html
@@ -17,6 +17,16 @@
 				}
 			},
 
+			min: {
+				type: Number,
+				value: null
+			},
+
+			max: {
+				type: Number,
+				value: null
+			},
+
 			_filterInput: {
 				type: Object,
 				value() {
@@ -31,7 +41,7 @@
 
 			_limit: {
 				type: Object,
-				computed: '_computeLimit(_range, _filterInput.*)'
+				computed: '_computeLimit(_range, _filterInput.*, min, max)'
 			},
 
 			_tooltip: {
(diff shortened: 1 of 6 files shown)


cosmoz-omnitable-column-range-behavior.html
```diff
@@ -0,0 +1,214 @@
...
+				acc.max = this.limitNumbers(Math.max, acc.max, this.getComparableValue(item));
+				acc.min = this.limitNumbers(Math.min, acc.min, this.getComparableValue(item));
+				return acc;
+			}, {});
+		},
+
+		_computeLimit(range, inputChange) {
+			const input = inputChange.base;
+			return {
+				fromMin: range.min,
```
nomego: This seems like it can only be the lowest number in `values` ?
We also need support for manually setting the `min` and `max` on the column, overriding the calculated min/max.
(Like we did previously with _getMinMax(_possibleAmountRange.min, min.*, 0)

cosmoz-omnitable-column-range-behavior.html
```diff
@@ -0,0 +1,214 @@
...
+		_computeTooltip: function (title, text) {
+			if (text == null) {
+				return title;
+			}
+			return `${title}: ${text}`;
+		},
+
+
+		_filterInputChanged: function () {
+			this.debounce('_updateFilter', this._updateFilter, 600);
```
nomego: This is probably a good default but please consider https://github.com/Neovici/cosmoz-omnitable/issues/60 if you haven't

cosmoz-omnitable-column-range-behavior.html
nomego: Ah I see that you do it later on. Why?

cosmoz-omnitable-column.html
```diff
@@ -24,7 +24,18 @@
 
 			behaviors: [
 				Cosmoz.OmnitableColumnBehavior
-			]
+			],
+			_serializeFilter: function (filter = this.filter) {
```
nomego: Wouldn't we need to consider URL encoding for this column at least?

megheaiulian: I've moved the url encoding function _serializeFilter and _deserializeFilter from `cosmoz-omnitable` to `OmnitableColumnBehavior`. The basic column needed to overwrite those because it works with strings.

cosmoz-omnitable-column-range-behavior.html
megheaiulian: I will add back support for `min`,`max` in the new behavior.

megheaiulian: Not sure what you mean with doing this later on ?

PR #108, opened 2017-12-12
Range columns tests

### LINKED ISSUE #104: Add tests and docs for RangeColumnBehavior
Add tests and docs for `RangeColumnBehavior` and columns using it:
* number
* amount
* date
* datetime
* time

programmer4web:
Add tests for range based columns.

fixes #104

### DIFF
--- a/test/range-date.html
+++ b/test/range-date.html
@@ -1,0 +1,374 @@
+<!doctype html>
+<html>
+<head>
+	<title>cosmoz-omnitable range tests for date, datetime, time</title>
+	<meta charset="UTF-8">
+	<meta name="viewport" content="width=device-width, minimum-scale=1.0, initial-scale=1.0, user-scalable=yes">
+
+	<script src="../../webcomponentsjs/webcomponents-lite.js"></script>
+	<script src="../../web-component-tester/browser.js"></script>
+	<script src="../../test-fixture/test-fixture-mocha.js"></script>
+
+	<link rel="import" href="../../test-fixture/test-fixture.html">
+	<link rel="import" href="../../iron-test-helpers/iron-test-helpers.html">
+
+	<link rel="import" href="../cosmoz-omnitable.html">
+	<link rel="import" href="../cosmoz-omnitable-columns.html">
+	<link rel="import" href="../demo/table-demo-behavior.html">
+</head>
+<body>
+	<test-fixture id="range">
+		<template>
+			<cosmoz-omnitable id="omnitable" class="flex">
+				<cosmoz-omnitable-column-number title="Age" name="age" value-path="age">
+				</cosmoz-omnitable-column-number>
+				<cosmoz-omnitable-column-amount title="Amount" name="amount" value-path="amount">
+				</cosmoz-omnitable-column-amount>
+				<cosmoz-omnitable-column-date title="Date" name="date" value-path="date" locale="en-US" editable>
+				</cosmoz-omnitable-column-date>
+				<cosmoz-omnitable-column-datetime title="Datetime" name="datetime" value-path="date" editable>
+				</cosmoz-omnitable-column-datetime>
+				<cosmoz-omnitable-column-time title="Time" name="time" value-path="time" editable>
+				</cosmoz-omnitable-column-time>
+			</cosmoz-omnitable>
+		</template>
+	</test-fixture>
+
+	<script>
+	/*global sinon chai flush */
+	(function () {
+		'use strict';
+		const data = [
+			{ age: 17,
+				amount: { amount: '12.4', currency: 'USD' },
+				date: '2023-03-21T00:00:00Z',
+				time: ''
+			},
+			{ amount: { amount: 2 },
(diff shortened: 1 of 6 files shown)


nomego:
What's left to drop the `wip` more than a rebase?

test/range-date.html
```diff
@@ -0,0 +1,319 @@
...
+				assert.typeOf(column.getComparableValue(data[0].date), 'number');
+				assert.typeOf(column.getComparableValue(data[1].date), 'number');
+				assert.typeOf(column.getComparableValue(data[2].date), 'number');
+				done();
+			});
+
+			test('getString returns formated date', function (done) {
+				assert.equal(column.getString(), '');
+				assert.isString(column.getString(data[0], 'date'));
+				assert.equal(column.getString(data[1]), '3/18/2015');
```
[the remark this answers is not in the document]
programmer4web: this is fixed as suggested. In Commit 'ISO 8601 strings with UTC time' at line 27:
b8115de84dbf22409a76c4aac1de156bc56f5928

programmer4web:
This pull request is ready for Review.

test/range-date.html
```diff
@@ -0,0 +1,337 @@
...
+				assert.equal(column.getString(data[3], 'date'), '6/24/1999');
+				assert.equal(column.getString(data[4]), '8/27/2016');
+				done();
+			});
+
+			test('_toLocalISOString returns formated date', function (done) {
+				assert.equal(column._toLocalISOString(new Date('2016-08-27')), '2016-08-27T03:00:00.000');
```
plequang: This works only if current timezone is EET which means if I want to run this test locally in a browser, I have to change my computer timezone to EET.
Also, will it continue to work when this timezone use daylight saving time ?

megheaiulian: Some different local timezone must be used.
Otherwise tests are irrelevant.
In travis.yml a timezone is forced and I think we are able to add that also in the yarn script but it depends also on the operating system.

megheaiulian: Tests should work with daylight saving too.

cosmoz-omnitable-column-date.html
```diff
@@ -52,7 +52,21 @@ <h3 style="margin: 0;">[[ title ]]</h3>
...
+			 * Get comparable number from date,
+			 * sets zero hours, minutes, seconds, milliseconds - this fixes Neovici/cosmoz-omnitable/issues/118
+			 * @returns {Number|void} Valid value or void
+			 */
+			getComparableValue() {
+				const value = Cosmoz.RangeColumnBehavior.getComparableValue.apply(this, arguments);
+				if (value == null) {
+					return;
+				}
+				return this.toNumber(new Date(value).setUTCHours(0, 0, 0, 0));
```
nomego: This is wrong. All timestamps at midnight east of UTC/GMT will be subtracted one day.
Another approach is needed, probably making sure we get the value of the filter max date with a "23:59:59" time added before doing getTime()

test/range-date.html
```diff
@@ -0,0 +1,367 @@
...
+				Polymer.Base.async(function () {
+					let fn = column.getFilterFn();
+					assert.isTrue(fn(data[5]));
+
+					done();
+				}, 650);
+			});
+
+			test('filtering on specific date with datetime source same day test', function (done) {
+				// date: '2017-12-22T23:00:00Z'
```
nomego: If this is run in Bucharest TZ, the test should actually be for 2017-12-23, not 22.

programmer4web: test updated in commit 'date tests small update' at line 193: 589ed7607bb6c34af850800e87666bd03899afe6

programmer4web:
I have rebased the tests.

cosmoz-omnitable-column-range-behavior.html
```diff
@@ -244,28 +270,28 @@
...
 			});
 			this.flushDebouncer('_limitInput');
 		},
 
 		_updateFilter() {
 			const input = this._filterInput,
 				filter = this.filter,
-				min = this._fromInputString(input.min),
-				max = this._fromInputString(input.max);
+				min = this._fromInputString(input.min, 'min'),
```
nomego: Hmm isn't this needed for the previously merged #120 ?

cosmoz-omnitable-column-range-behavior.html
```diff
@@ -277,11 +303,16 @@
 		},
 
 		_filterChanged(change) {
+			if (this._filterInput == null) {
+				return;
+			}
 			const input = this._filterInput,
-				filter = change.base;
+				filter = change.base,
+				min = this._fromInputString(input.min, 'min'),
```
nomego: Also this for #120 ?

test/range-date.html
programmer4web: @plequang The different timezone is now set with a environment variable in yarn test script so it should also work locally.

programmer4web:
@nomego Moved fixes and docs into #121. In this PR only tests are kept.

PR #109, opened 2017-12-13
Expose omnitable visibility

nomego:


cosmoz-omnitable.js
```diff
@@ -187,6 +187,12 @@
 				notify: true
 			},
 
+			visible: {
```
plequang: Would be safer to make it `readOnly`

PR #114, opened 2017-12-15
omnitable render on visible

nomego:
The most expensive operation when rendering views right now is the
omnitable setup. When a data-nav renders a view with 2-3 omnitables
each, hidden away behind tabs, it doesn't make sense that one of the
biggest rendering chunks is _updateColumns.

This PR tries to reduce omnitable setup overhead.

The visibility control is done earlier, in _updateColumns rather than
_adjustColumns, and will prevent _updateColumns to run if not visible.

Also, resize will trigger visibility re-evaluation.

When omnitable turns visible, _updateColumns will not have run, so we
make sure to trigger it then.

Finally, an immediate _updateColumns in _columnObserver will prevent it
to run twice when loaded visibly.

### DIFF
--- a/cosmoz-omnitable.js
+++ b/cosmoz-omnitable.js
@@ -268,7 +268,7 @@
 
 				this._setColumnValues(addedColumns);
 
-				this._updateColumns();
+				this._debounceUpdateColumns();
 			});
 			this.$.groupedList.scrollTarget = this.$.scroller;
 			this.listen(this,  'cosmoz-column-hidden-changed', '_debounceUpdateColumns');
@@ -393,7 +393,7 @@
 		},
 
 		_debounceUpdateColumns() {
-			this.debounce('updateColumns', this._updateColumns);
+			this.debounce('updateColumns', this._updateColumns, 1);
 		},
 
 		_updateColumns() {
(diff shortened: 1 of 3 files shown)


cosmoz-omnitable.js
```diff
@@ -267,7 +268,7 @@
 
 				this._setColumnValues(addedColumns);
 
-				this._debounceUpdateColumns();
+				this._updateColumns();
```
plequang: When the omnitable is immediately visible, a first `_updateColumn` is run by `onResize` , because an element implementing `iron-resizable` will get a first `iron-resize` event when it's attached. 
The `updateColumns` call from `observeNodes` will occur after that.

If you want to be sure to run `_updateColumns` only once, you should probably always debounce, with a delay greater than 0.

nomego:
@JaySunSyn helped out to fix the tests, any feedback now @plequang @megheaiulian ?

JaySunSyn:
My PR https://github.com/Neovici/cosmoz-omnitable/pull/119 to this PR failed building. When merged, building succeeded. Maybe someone knows how to pull out the luck factor here...

nomego:
Closing since this has been implemented through #132

PR #120, opened 2018-01-04
Filtering date fix

programmer4web:
fixes Neovici/cosmoz-omnitable/issues/118

### DIFF
--- a/cosmoz-omnitable-column-date.html
+++ b/cosmoz-omnitable-column-date.html
@@ -52,7 +52,25 @@
 				Cosmoz.OmnitableColumnBehavior,
 				Cosmoz.TranslatableBehavior,
 				Cosmoz.DateColumnBehavior
-			]
+			],
+
+			_fromInputString(value, property) {
+				let date = this.toDate(value);
+				if (date == null) {
+					return;
+				}
+				if (property === 'min') {
+					date.setUTCHours(0, 0, 0, 0);
+				}
+				if (property === 'max') {
+					date.setUTCHours(23, 59, 59);
+				}
+				return date;
+			},
+
+			_fromHashString(value, property) {
+				return this._fromInputString(value, property);
+			}
 		});
 
 	</script>
--- a/cosmoz-omnitable-column-date.html
+++ b/cosmoz-omnitable-column-date.html
@@ -66,10 +66,6 @@
 					date.setHours(23, 59, 59);
 				}
 				return date;
-			},
-
-			_fromHashString(value, property) {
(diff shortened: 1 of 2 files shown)


cosmoz-omnitable-column-date.html
```diff
@@ -52,7 +52,25 @@ <h3 style="margin: 0;">[[ title ]]</h3>
 				Cosmoz.OmnitableColumnBehavior,
 				Cosmoz.TranslatableBehavior,
 				Cosmoz.DateColumnBehavior
-			]
+			],
+
+			_fromInputString(value, property) {
+				let date = this.toDate(value);
```
nomego: `const` ! :)

cosmoz-omnitable-column-date.html
```diff
@@ -52,7 +52,25 @@ <h3 style="margin: 0;">[[ title ]]</h3>
...
+				if (date == null) {
+					return;
+				}
+				if (property === 'min') {
+					date.setUTCHours(0, 0, 0, 0);
```
nomego: `date` will become a local date with `toDate()`, so to make this work as intended, we should not use the UTC-methods but the `setHours()` and similar.
Please ensure that the test from https://github.com/Neovici/cosmoz-omnitable/pull/108/files#diff-4775d6b705fe614d00ac79e1629929c4R190 is passing (locally).

cosmoz-omnitable-column-date.html
```diff
@@ -52,7 +52,25 @@ <h3 style="margin: 0;">[[ title ]]</h3>
 				Cosmoz.OmnitableColumnBehavior,
 				Cosmoz.TranslatableBehavior,
 				Cosmoz.DateColumnBehavior
-			]
+			],
+
+			_fromInputString(value, property) {
```
nomego: This doesn't have to be a helper method on its own at this point, maybe rather just call it `_fromHashString` ?

programmer4web: _fromInputString is an function originally defined in range behavior.

nomego: Ok but then we don't need to override `_fromHashString` rather since it's the same as the one inherited?

nomego: Ah right it was a second argument, didn't see that!
Anyway gonna merge this now before I mess it up more :)

programmer4web: I have already updated the original _fromHashString with the second argument. :)

Issue #123, opened 2018-01-05
Download XLSX 2nd time doesn't work


MickeBerggren:
Nothing happens when trying to Download a 2nd XLSX.

nomego:
Fixed @MickeBerggren ?

MickeBerggren:
Yes

PR #125, opened 2018-01-05
Used on-click instead of on-action and click tests

programmer4web:
fixes: second click does not download file. https://github.com/Neovici/cosmoz-omnitable/issues/123


nomego:
We should drop using on-tap (https://www.polymer-project.org/2.0/docs/devguide/gesture-events).
This change should be fine for the download menu but similar behaviors should exist for action buttons/menu items in bottom-bar.
We should add tests for the action handling and refactor.

Issue #116, opened 2017-12-21
Replace JaySunSyn/datetime-local with Neovici/cosmoz-datetime-input


plequang:
Currently, we are https://github.com/JaySunSyn/datetime-local component to render datetime values.
We should replace it with Neovici/cosmoz-datetime-input

JaySunSyn:
can be closed

nomego:
Did you remove your comment? Looking at bower.json we still use your repo.

JaySunSyn:
Yes, removed it. Thought that already has been done..

JaySunSyn:
Oh.. Didn't see this one. https://github.com/Neovici/cosmoz-omnitable/pull/124

PR #126, opened 2018-01-08
Replaced datetime-local with cosmoz-datetime-input

JaySunSyn:
See https://github.com/Neovici/cosmoz-omnitable/issues/116

Please also consider merging https://github.com/Neovici/cosmoz-datetime-input/pull/3


plequang:
Merged #124 which was first

PR #131, opened 2018-01-10
Improve range columns limiting

megheaiulian:
Improve range columns limiting

### DIFF
--- a/cosmoz-omnitable-column-date.html
+++ b/cosmoz-omnitable-column-date.html
@@ -88,6 +88,19 @@
 				localDate.setHours(0, 0, 0, 0);
 				return localDate;
 			}
+			},
+
+			_filterInputChanged(change) {
+				const path = change.path.split('.')[1],
+					value = path && change.value;
+
+				if(value && value.match(/^0+/)){
+					this.cancelDebouncer('_limitInput');
+					return;
+				}
+				Cosmoz.RangeColumnBehavior._filterInputChanged.call(this, change);
+			},
+
 		});
 
 	</script>
--- a/cosmoz-omnitable-column-date.html
+++ b/cosmoz-omnitable-column-date.html
@@ -87,19 +87,18 @@
 				const localDate = this.toValue(this._toLocalISOString(date));
 				localDate.setHours(0, 0, 0, 0);
 				return localDate;
-			}
 			},
 
 			_filterInputChanged(change) {
 				const path = change.path.split('.')[1],
 					value = path && change.value;
 
-				if(value && value.match(/^0+/)){
+				if (value && value.match(/^0+/)) {
 					this.cancelDebouncer('_limitInput');
 					return;
 				}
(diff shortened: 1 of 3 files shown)


programmer4web:
@nomego This is ready for review.

cosmoz-omnitable-column-range-behavior.html
```diff
@@ -271,22 +272,27 @@
...
-				min = this._fromInputString(input.min, 'min'),
-				max = this._fromInputString(input.max, 'max'),
-				lMin = this.toValue(min, limit.fromMax, Math.min),
-				lMax = this.toValue(max, limit.toMin, Math.max);
-
-			if (this.getComparableValue(lMin) === this.getComparableValue(min)
-				&& this.getComparableValue(lMax) === this.getComparableValue(max)) {
-				return this._updateFilter();
+				limit = this._limit;
+
```
nomego: There are quite a few things that feel sub-optimal here. (Not just from this specific PR)
We limit the input after a 600ms debounce, which I guess could make the value change once, causing any outer usage/listeners to trigger OTS-search, then the input limiter runs, adjusting the input value, triggering another OTS search?
Two other "smells" are also present, the responsibility of running "updateFilter" after "_limitInput" lies with the "_limitInput" function - can we control this in another manner "outside" with observer "dependencies"? This causes us to need the additional complexity, instead of `if (path == null) { return; }` and the equivalent for `value`. We could do `if (path == null) { return this._updateFilter(); }` but it's risky for maintainability.

cosmoz-omnitable-column-date.html
```diff
@@ -87,7 +87,19 @@ <h3 style="margin: 0;">[[ title ]]</h3>
 				const localDate = this.toValue(this._toLocalISOString(date));
 				localDate.setHours(0, 0, 0, 0);
 				return localDate;
+			},
+
+			_filterInputChanged(change) {
+				const path = change.path.split('.')[1],
+					value = path && change.value;
+
+				if (value && value.match(/^0+/)) {
```
nomego: I think we need a comment for this regex (I think most regex should have one).
Why does value need to start with one or more zeroes? Why should `0.1234` be handled differently than `1.1234`?

programmer4web: I think it was an problem with filtering dates when an year is typed for example 2017 the input value will be 0002 then 0020, 0201 then 2017 for the cases when the value started with 0 then the search was triggered.

nomego: Ah right it's the date column, my bad.
Anyway we should add a comment for this and/or design the condition differently.
So what you're really after is detecting invalid dates?

programmer4web: I will add the comment in the code and yes dates starting with zero were considered as invalid because these prevented the typing of any other year in combination with the filtering code (limiting to min or max of the input) ( the initial value for year is 0000 ).

megheaiulian:
@nomego Updated the PR. Please have a look again.

PR #132, opened 2018-01-10
Render on visible (with tests)

megheaiulian:
PR #114 had problems with tests.
Fixed tests and related misc issues. 

The main problem was that a resize was needed in order to make sure everything is correct when checking if params from hash are reflected in omnitable properties.


megheaiulian:
Updated tests to use events instead of timeouts. Mainly when testing filter input changes we listen for `filter-changed` in tests.

PR #136, opened 2018-01-12
Add immediateDataInit property

megheaiulian:
Add `immediateDataInit` property to get columns, update from hash params and debounce filter event when omnitable is not visible

### DIFF
--- a/cosmoz-omnitable.js
+++ b/cosmoz-omnitable.js
@@ -227,6 +227,11 @@
 				computed: '_computeRouteHashKeyRule(hashParam)'
 			},
 
+			immediateDataInit: {
+				type: Boolean,
+				value: false
+			},
+
 			/**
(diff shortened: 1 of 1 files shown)


nomego:
I'm not sure this is the best way to go.
I think the ideal behavior would be to init omnitable every time but only do expensive dom/reflow/calc style stuff if/when visible. For huge data sets maybe rather have a `noDataInit` to defer that work.

megheaiulian:
@nomego I think we decided to not implement this anymore? Can we close the PR and related issue ?

nomego:
Sure, I think the visibility constraint to init omnitable avoids the biggest perf hit that this tried to solve.
If we want to add something like this I think it should be the reverse - to turn off init on load, for certain instances. Also we likely want to have a look a cosmoz-tab template support or similar before we do this.

Issue #133, opened 2018-01-11
XLSX shows faulty date and lacks article-ID


fredrikbernholm:
Downloaded XLSX (from omnitable) shows weird date format and lacks article-ID.

fredrikbernholm:
Can't seem to assign issues, but this one's for you @JaySunSyn

nomego:
Actually this is high prio so who of you three would like to have a look?

JaySunSyn:
@megheaiulian 
I've changed in `cosmoz-omnitable-column-range-behavior.html`:
```javascript
toXlsxValue: function (item, valuePath = this.valuePath) {
    const value = this.getComparableValue(item, valuePath);
    if (value == null) {
        return '';
    }
    return value;
},
```
to
```javascript
toXlsxValue: function (item, valuePath = this.valuePath) {
    const value = this.getString(item, valuePath);
    if (value == null) {
        return '';
    }
    return value;
},
```
is that fine for you? Do I miss something?

---

@nomego in the FE there was a value-path mismatch of the article id. 
`article.number` (frontend) vs. `articleNumber` (response from backend)
Which one is right? See https://github.com/Neovici/cosmoz-frontend/pull/97

nomego:
I heard something about XLSX wanting dates in an integer-kind of format (from @jokedst) and based on that we might want something like getComparableValue. But we also have the scenario where a datetime is the source of a date column and then we expect the value to be a date without time. We have to be careful when we do that conversion and right now we should probably do it in the toXlsxValue function.

JaySunSyn:
It seems like we are doing the integer conversion already in `nullxlsx`. (checks value instanceof Date). Need to check what happens with time & datetime ...

JaySunSyn:
I think we can solve it like this:
- datetime: Pass in JS Date object
- date: Pass in (time cleared) JS Date object
- time: Pass in formatted time string;

Regarding time: The alternative would be a time formatted cell in nullxlsx. The excel file would then only show the time value which would be much cleaner. When selecting the cell though, the date value would be visible. As the time column sets a random date when passing in a time-only string, we, unfortunately, cannot do it like this ...

nomego:
Sounds good but we can't just "clear" time, the datetime has to be evaluated to local time first before we strip it.

JaySunSyn:
Sorry, didn't see this. But I see it is already done.

PR #140, opened 2018-01-15
toXlsxValue date to local time

programmer4web:
fixes: https://github.com/Neovici/cosmoz-omnitable/issues/133


megheaiulian:
based on  #137

megheaiulian:
@programmer4web We should add tests for xlsx export to check for date, time, datetime, amount and number columns. For date columns we should test with dates in different timezones.

PR #149, opened 2018-01-26
Improve autocomplete serialize, deserialize

programmer4web:
_serializeFilter for example if valueProperty is id then in url will be only ids separated by ~
deserializeFilter splits ids and searches in column values for them and returns the items array.

Iulian: The problem is that when we set filter (from deserializeFilter) values might not be set if data property is not set so it can't set selection in paper-autocomplete-chips.

version 1:  Because of that I set the param from url as _pendingFilterString and next time values are set it will update filter.

version 2: ( without flag ) set an array of values in paper-autocomplete-chips. [WIP]

### DIFF
--- a/cosmoz-omnitable-column-autocomplete.html
+++ b/cosmoz-omnitable-column-autocomplete.html
@@ -168,6 +168,34 @@
 					}
 					return true;
 				});
+			},
+
+			_serializeFilter: function (obj = this.filter) {
+				if (!(Array.isArray(obj) && obj.length > 0)) {
+					return null;
+				}
+				return obj.map(item => {
+					const property = this.valueProperty;
+					if (!property) {
+						return item;
+					}
+					return this.get(property, item);
+				}).join('~');
+			},
+
+			_deserializeFilter: function (obj) {
+				if (obj == null || obj  === '') {
+					return null;
+				}
+				return obj.split('~').map(item => this.values.find(v => {
+					const property = this.valueProperty;
+					if (!property) {
+						const vv = typeof v === 'number' ? v.toString() : v;
+						const ii = typeof item === 'number' ? item.toString() : item;
+						return vv === ii;
+					}
+					return this.get(property, v) === item;
+				}));
 			}
 		});
 	</script>
--- a/cosmoz-omnitable-column-autocomplete.html
+++ b/cosmoz-omnitable-column-autocomplete.html
@@ -92,6 +92,9 @@
 					notify: true
 				},
 			},
+			observers: [
+				'_valuesForHashChanged(values.*)'
+			],
 
 			getComparableValue: function (item, valuePath = this.valuePath) {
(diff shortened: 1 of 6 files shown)


cosmoz-omnitable-column-autocomplete.html
```diff
@@ -158,6 +158,34 @@
 					}
 					return true;
 				});
+			},
+
+			_serializeFilter: function (obj = this.filter) {
```
nomego: - Use es6 syntax and skip `: function `
- Why `obj` (more generic, harder to follow) and not something like just `filter` ?

(same below)

cosmoz-omnitable-column-autocomplete.html
```diff
@@ -158,6 +158,34 @@
...
+			},
+
+			_deserializeFilter: function (obj) {
+				if (obj == null || obj  === '') {
+					return null;
+				}
+				return obj.split('~').map(item => this.values.find(v => {
+					const property = this.valueProperty;
+					if (!property) {
+						const vv = typeof v === 'number' ? v.toString() : v;
```
nomego: Why is this needed? Numbers should be able to compare directly, `3 === 3`

cosmoz-omnitable-column-autocomplete.html
megheaiulian: @nomego Fixed that.

cosmoz-omnitable-column-autocomplete.html
megheaiulian: @nomego I changed to a `_normalizeValue` function because numbers from hash params are strings sometimes. If possible we convert to numbers.

cosmoz-omnitable-column-autocomplete.html
```diff
@@ -168,7 +171,54 @@
...
+					.map(item => this.values
+						.find(v => this._normalizeValue(property ? this.get(property, v) : v) === this._normalizeValue(item))
+					).filter(i => i != null);
+			},
+
+			_valuesForHashChanged() {
+				if (!Array.isArray(this.values) || this._pendingFilterString) {
+					return;
+				}
+				this.setFilterFromHash(this._pendingFilterString);
```
nomego: `_pendingFilterString` can only be falsy? Intended? Where does that property even come from? New in this PR?

megheaiulian: @nomego The problem  is that when we set filter (from `deserializeFilter`) `values` might not be set if `data` property is not set so it can't set selection in paper-autocomplete-chips.
Because of that I set the param from url as `_pendingFilterString` and next time values are set it will update filter.

megheaiulian: Another solution would be to able to set an array of values in `cosmoz-autocomplete-chips`.

Issue #151, opened 2018-01-31
Clear buttons are always visible with polymer 2.0


programmer4web:
Filter clear buttons, input clear icons in columns headers are always visible.

PR #152, opened 2018-01-31
Fix for clear buttons always visible in polymer 2

programmer4web:
fixes https://github.com/Neovici/cosmoz-omnitable/issues/151
In cosmoz-omnitable-column the close iron-icon with hidden attribute was visible because of suffix ::slotted(*) { display: inline-block;  Used the existing ui-helper: cosmoz-clear-button fixing css for visible attribute. Added light attribute to keep the original css of iron-icon close.

### DIFF
--- a/cosmoz-omnitable-column.html
+++ b/cosmoz-omnitable-column.html
@@ -16,7 +16,7 @@
 
 		<template class="header">
 			<paper-input label="[[ title ]]" value="{{ filter }}" focused="{{ headerFocused }}">
-				<cosmoz-clear-button suffix slot="suffix" visible="[[ hasFilter(filter.*) ]]" filter=true on-click="resetFilter"></cosmoz-clear-button>
+				<cosmoz-clear-button suffix slot="suffix" visible="[[ hasFilter(filter.*) ]]" light on-click="resetFilter"></cosmoz-clear-button>
 			</paper-input>
 		</template>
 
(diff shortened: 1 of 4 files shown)


programmer4web:
@plequang This is ready for review and needed in cosmoz-frontend  with polymer 2.

cosmoz-omnitable-column.html
```diff
@@ -15,7 +16,7 @@
 
 		<template class="header">
 			<paper-input label="[[ title ]]" value="{{ filter }}" focused="{{ headerFocused }}">
-				<iron-icon icon="clear" suffix slot="suffix" hidden$="[[ !filter.length ]]" style="cursor: pointer;" on-click="resetFilter"></iron-icon>
+				<cosmoz-clear-button suffix slot="suffix" visible="[[ hasFilter(filter.*) ]]" light on-click="resetFilter"></cosmoz-clear-button>
```
plequang: Found an strange bug with `hasFilter(filter.*)` here.

If you try the full-demo, and type a letter in the `Name` header field, `hasFilter`function is called.

In the code of `hasFilter`, we have:
```
hasFilter() {
   const filter = this.filter,
```
At that time, `this.filter` is empty, but checking the `arguments` variable shows the typed letter.
As if the `visible="[[ hasFilter(filter.*) ]]"` binding has been evaluated before the `filter` property has been set.

So because of that, `hasFilter` returns `false`, and the clear icon is not shown.
If you remove the letter, pressing backspace, then the clear icon is shown.

😿

programmer4web: @plequang Fixed hasFilter with commit 659c0a452300daa1a111ab41dfaa19089006931f

PR #155, opened 2018-02-19
Fit dropdown into omnitable

### LINKED ISSUE #58: Fit headers and footer dropdowns into the omnitable
# Description
Some column headers are using dropdowns by mean of https://www.webcomponents.org/element/PolymerElements/paper-dropdown-menu.

The paper-dropdown-menu's use default `vertical-align=top`, and `horizontal-align` is `left` for the first column, `right` for others.

We should rather use `dynamic-align` with `fit-into` properties from `iron-dropdown` in order to have the dropdown align and fit into the table dynamically.

However, this will be possible only when `fit-into` will be exposed by `paper-menu-button` and `paper-dropdown-button`, as explained in these issues:
* https://github.com/PolymerElements/paper-menu-button/issues/108
* https://github.com/PolymerElements/paper-dropdown-menu/issues/218

Corresponding PRs:
* https://github.com/PolymerElements/paper-menu-button/pull/130
* https://github.com/PolymerElements/paper-dropdown-menu/pull/266

When these issues are fixed, we will have to expose a reference to the omnitable in the header template instances, and bind `fit-into` to that property in `paper-dropdown-menu`s.

megheaiulian:
On attach set `paperDropdownMenu.$.menuButton.$.dropdown.fitInto` to `this`

Fixes #58

### DIFF
--- a/cosmoz-omnitable.js
+++ b/cosmoz-omnitable.js
@@ -274,6 +274,8 @@
 			});
 			this.$.groupedList.scrollTarget = this.$.scroller;
 			this.listen(this,  'cosmoz-column-hidden-changed', '_debounceUpdateColumns');
+
+			this._fitDropdowns();
 		},
 
 		detached: function () {
@@ -1220,6 +1222,15 @@
 			}
 
 			this.set(path, serialized === undefined ? null : serialized);
+		},
+
+		_fitDropdowns() {
+			const dropdowns = [this.$.groupOnSelector, this.$.sortOnSelector];
+
+			dropdowns.forEach(dropdown => {
+				dropdown.$.menuButton.$.dropdown.fitInto = this;
+			});
 		}
+
 	});
 }());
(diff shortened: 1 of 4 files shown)


cosmoz-omnitable.js
```diff
@@ -1220,6 +1222,17 @@
 			}
 
 			this.set(path, serialized === undefined ? null : serialized);
+		},
+
+		_fitDropdowns() {
+			const menuButtons = [this.$.groupOnSelector, this.$.sortOnSelector]
```
nomego: Beautiful! You don't even need `menuButtons` :D

megheaiulian:
@nomego Should the same be done for dropdowns in headers? There we must hook into the getTemplateInstance function somehow and grab any dropdowns that we can find.

megheaiulian:
@nomego Will add tests also.

PR #157, opened 2018-03-02
Do not override sortOn and groupOn properties on columns with valuePath when they are defined

plequang:
* Do not update `sortOn` or `groupOn` property of columns with `valuePath` if they have been set
* Added test to show issue with column's groupOn and sortOn properties
* Attempt to fix weird timeout when running tests

### DIFF
--- a/cosmoz-omnitable-column-behavior.html
+++ b/cosmoz-omnitable-column-behavior.html
@@ -94,12 +94,12 @@
 
 			sortOn: {
 				type: String,
-				value: '' // need a default value for compatibility with Polymer 1.x, to trigger _pathsChanged observer
+				value: null // need a default value for compatibility with Polymer 1.x, to trigger _pathsChanged observer
 			},
 
 			groupOn: {
 				type: String,
-				value: '' // need a default value for compatibility with Polymer 1.x, to trigger _pathsChanged observer
+				value: null // need a default value for compatibility with Polymer 1.x, to trigger _pathsChanged observer
 			},
 
 			/**
@@ -315,15 +315,15 @@
 		},
 
 		_pathsChanged(valuePath, groupOn, sortOn) {
-			if (valuePath === undefined || valuePath === null) {
+			if (valuePath == null) {
 				return;
 			}
 
-			if (groupOn === undefined || groupOn === '') {
+			if (groupOn == null) {
 				this.set('groupOn', valuePath);
 			}
 
-			if (sortOn === undefined || sortOn === '') {
+			if (sortOn == null) {
 				this.set('sortOn', valuePath);
 			}
 		},
--- a/cosmoz-omnitable-column-behavior.html
+++ b/cosmoz-omnitable-column-behavior.html
@@ -17,8 +17,7 @@
 			},
 
 			valuePath: {
-				type: String,
-				observer: 'valuePathChanged'
+				type: String
 			},
 
 			/**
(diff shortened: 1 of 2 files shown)


test/basic.html
```diff
@@ -39,7 +47,6 @@
 				</cosmoz-omnitable-column-date>
 				<cosmoz-omnitable-column-date id="date3" value-path="date">
 				</cosmoz-omnitable-column-date>
-			</cosmoz-omnitable>
```
nomego: Accidental drop ?

cosmoz-omnitable-column-behavior.html
```diff
@@ -312,15 +314,17 @@
...
-		valuePathChanged: function (newPath, oldPath) {
-			if (newPath === undefined || newPath === null) {
+		_pathsChanged: function (valuePath, groupOn, sortOn) {
+			if (valuePath === undefined || valuePath === null) {
 				return;
 			}
-			if (this.groupOn === oldPath || this.groupOn === undefined) {
-				this.set('groupOn', newPath);
+
+			if (groupOn === undefined || groupOn === null || groupOn === '') {
```
nomego: == null ?

cosmoz-omnitable-column-behavior.html
```diff
@@ -312,15 +314,17 @@
 			return this.get(valuePath, item);
 		},
 
-		valuePathChanged: function (newPath, oldPath) {
-			if (newPath === undefined || newPath === null) {
+		_pathsChanged: function (valuePath, groupOn, sortOn) {
```
nomego: `function` ?! :)

test/basic.html
plequang: Yes

cosmoz-omnitable-column-behavior.html
plequang: True, not needed

cosmoz-omnitable-column-behavior.html
plequang: Ok Ok

cosmoz-omnitable-column-behavior.html
```diff
@@ -312,15 +314,17 @@
 			return this.get(valuePath, item);
 		},
 
-		valuePathChanged: function (newPath, oldPath) {
-			if (newPath === undefined || newPath === null) {
+		_pathsChanged(valuePath, groupOn, sortOn) {
+			if (valuePath === undefined || valuePath === null) {
```
nomego: :heavy_plus_sign: for `== null` :)

cosmoz-omnitable-column-behavior.html
```diff
@@ -312,15 +314,17 @@
...
 				return;
 			}
-			if (this.groupOn === oldPath || this.groupOn === undefined) {
-				this.set('groupOn', newPath);
+
+			if (groupOn === undefined || groupOn === '') {
```
nomego: Can we really check for `=== ''` ? Wouldn't this prevent issues when not grouping, like re-enabling grouping on `valuePath` if `sortOn` changes?

test/basic.html
nomego: .. so should it be added back? :)

cosmoz-omnitable-column-behavior.html
plequang: True.

Current code hides groupOn option for a column with this attribute
```html
hidden$="[[ !column.groupOn ]]"
```
(and the same for sortOn)

So I'll drop the test.

Though I'm wondering if it would be more "correct" to have `no-group-on`, `no-sort-on`.

Not sure actually.

PR #159, opened 2018-03-02
refs #158 Added a new `titlePath` property to columns

### LINKED ISSUE #158: When binding an object to a column, `[object Object]` is set in the cell wrapping div for this column
# Description

Given the following column definition

```
<cosmoz-omnitable-column title="Status" value-path="status" name="status">
   <template class="cell">
      <span>[[  item.status.description ]]</span>
   </template>
````

and the following table data:
```
[
   {
      "status": {
         "description":  "some status description",
         "value":  1
      }
   },
....
```

then the generated `<div>` element that contains the cell will look like:
```
<div title="[object Object]" class="itemRow-cell default-cell
```` 

When rendering the title, we don't expect `valuePath` to refer to an object:

https://github.com/Neovici/cosmoz-omnitable/blob/7b13ec80f470e12c5da93a898f4957c7554a9061/cosmoz-omnitable-item-row.html#L125-L127

### PR DESCRIPTION
**Pull request by plequang**
Added a new `titlePath` property to columns to fix #158

### DIFF
--- a/cosmoz-omnitable-column-behavior.html
+++ b/cosmoz-omnitable-column-behavior.html
@@ -22,6 +22,16 @@
 			},
 
 			/**
+			 * Path to a property of a table's item that will be used as the title of the cell.
+			 * The title of the cell is a text usually displayed in a 'tool tip' popup when the mouse is over it.
+			 * This property should be set when valuePath reference an object, to prevent the title of the cell
+			 * to be [object Object].
+			 */
+			titlePath: {
+				type: String
+			},
+
+			/**
 			 * If the column should be disabled until enabled with enabledColumns
 			 */
 			disabled: {
--- a/cosmoz-omnitable-item-row.html
+++ b/cosmoz-omnitable-item-row.html
@@ -123,7 +123,7 @@
 		},
 
 		_getCellTitle: function (column, item) {
-			return column && column.getString(item, column.valuePath);
+			return column && column.getString(item, column.titlePath || column.valuePath);
 		},
 	});
 }());

### CODE REVIEW

**Comment by plequang** [comment]
This concerns columns bound to a `status` object, that has several properties
```
{
   "name": "status name",
   "description": "(translated) status description",
   "severity": "some value"
}
```
Some columns bound to this type of object use custom template, so I'm not sure paper-autocomplete-chips would make it.

However, I realize now that binding an object in a `cosmoz-omnitable-column` does not work really well.  For example, searching is not working.

I guess I should rather bind to the `description` property of the status.

Maybe this `title-path` is not needed after all.

**Comment by programmer4web** [comment]
<cosmoz-omnitable-column-autocomplete title="[[ _('Status', t) ]]" value-path="status" text-property="description" name="status" priority="-3">

**Comment by plequang** [comment]
@programmer4web  actually I'm talking about the `title` attribute that is set on every cell.

In the demo if you try to add the following column:
```
<cosmoz-omnitable-column title="Test" name="test" value-path="sub" sort-on="sub.subProp" group-on="sub.subProp" flex="5">
   <template class="cell">
      <span>[[item.sub.subProp]]</span>
   </template>
</cosmoz-omnitable-column>
```

When you hover a cell of this column, this is what you will see:
image

programmer4web:
Thanks for details.

nomego:
Well I think we're all talking about the same thing.
If you skip the custom template and specify `text-property="subProp"` to autocomplete ?

plequang:
@nomego Actually you're right, autocomplete column makes it, even with the custom template (there is a link on that column).

I think this PR can be closed.

PR #171, opened 2018-04-13
fixes #170 filter of bolean column not set when false value is selected.

### LINKED ISSUE #170: Bug, filtering boolean column on No does not filter out Yes rows
Selecting No in column header in a boolean column does not filter out Yes alternatives.

To reproduce the problem:

1. Login to beta
2. Go to Administration -> List users
3. In the Active column, select No
4. Rows with Yes still are in the table

This may be related to or the same as https://github.com/Neovici/cosmoz-omnitable/issues/169 but in differ from that issue this table responds to column header filter changes and updates its content.

plequang:


plequang:
@megheaiulian Could have look please, since @nomego might not have time to review this PR ?

cosmoz-omnitable-column-boolean.html
```diff
@@ -102,7 +102,11 @@
 					return;
 				}
 				this._filterChangedFromInput = true;
-				this.set('filter', textFilter && textFilter === 'true' ? true : null);
+				if (textFilter) {
+					this.set('filter', textFilter === 'true' ? true : false);
```
nomego: `? true : false` is unnecessary and I think we had a lint rule for it before?

PR #172, opened 2018-04-26
Check for null formatter in date columns

megheaiulian:
Check for null formatter in date columns.
Formatter can be null/undefined in 2.x when filter is set to value before the column is initialized.


megheaiulian:
@nomego `formatter` is not always passed to `renderValue` in data behavior and `this.formatter` is null/undefined (not computed yet).

PR #174, opened 2018-05-15
Set _methodHost  to the host property

megheaiulian:
Set _methodHost  to the host property to fix` method hasFilter not defined` warning.


nomego:
Could this have any other side effects?

megheaiulian:
This should not have any side effects. `_methodHost` describes the component on which the methods used in the template are called. It's weird that this is needed but I guess it's because some property (filter) is getting set very early. 
We should do a switch to Polymer.Templatize and this won't be needed anymore.

PR #188, opened 2018-06-08
Multi select a range with shift-select [issue-77]

### LINKED ISSUE #177: multi-select
Selecting an item in omnitable, then holding down shift and selecting another item further down the list, should also select all items in between

(See multi-select in Gmail)

JaySunSyn:
fixes #177 

Multi-selection with shift key pressed.

- works for grouped and non-grouped views

Please merge https://github.com/Neovici/cosmoz-grouped-list/pull/20 first and release version `1.0.3` of `cosmoz-grouped-list`. Otherwise build fails.


megheaiulian:
Implemented differently.

Issue #176, opened 2018-05-17
Filtering with hidden decimals


JockeCK:
In number columns we show a certain amount of fractions visually. Variable "minimumFractionDigits"

However the fractions are still there and can fool the user depending on how the filter is used.
Example:
46,768 is the number, lets say this is an currency amount and two fractions are shown.
That is as 46,76

The user sets both to and from filter as 46,76.

This results that at least not the row with the value 46,768 is not shown. It interprets the maximum allowed value as 46,760. 

So we need to take this into consideration in the range behavior. 

Top of my head solution would be to round the comparison value during comparisons according to "minimumFractionDigits". Or maybe just multiply both sides of the equation with an nice large number and call it a day.

Any thoughts @nomego ?

(RM #18323)

megheaiulian:
@nomego @JockeCK I think the best solution is to round according to `minimumFractionDigits` in the comparison function as suggested by @JockeCK.

nomego:
Sounds good, want to pr it?

programmer4web:
[WIP] writing an test to replicate. Then apply / implement fix.

PR #192, opened 2018-06-20
Number getComparableValue with maximumFractionDigits tests, fix.

programmer4web:
fixes: #176

1. Implemented Test: item with age: 46.768 and amount: 2581 should be in the filtered items when filtering from 46.76 to 46.76  Done.
2. In cosmoz-omnitable-column-number.html added getComparableValue and added the check for minimumFractionDigits. Value is truncated to the specified decimals (2 in the test). Done.
3. Rebased on latest changes from master. Done.
4. Replaced minimumFractionDigits with maximumFractionDigits in column and range test. Done.

5. Implement Test: If we filter out values between 46,75 - 46,76, we shouldn't show a value with 46,768 since it's outside the range. Done.

6. restore toFixed to round up instead of truncate to n digits. Done.
7. Implement test: But we should show 46,75xx. Done.
8. Verify if: However, if 46,768 is displayed as 46,76 then that's the error, it should be 46,77. Done.

### DIFF
--- a/cosmoz-omnitable-column-number.html
+++ b/cosmoz-omnitable-column-number.html
@@ -113,8 +113,9 @@
 					return;
 				}
 
-				const decimals = this.minimumFractionDigits;
+				const decimals = this.maximumFractionDigits;
 				if (decimals !== null) {
+					// truncate value to maximum decimals; https://github.com/Neovici/cosmoz-omnitable/issues/176
 					const d = Math.pow(10, decimals),
 						temp = parseInt(value * d, 10) / d;
 					return temp.toFixed(decimals);
--- a/cosmoz-omnitable-column-number.html
+++ b/cosmoz-omnitable-column-number.html
@@ -115,10 +115,7 @@
 
 				const decimals = this.maximumFractionDigits;
 				if (decimals !== null) {
-					// truncate value to maximum decimals; https://github.com/Neovici/cosmoz-omnitable/issues/176
-					const d = Math.pow(10, decimals),
-						temp = parseInt(value * d, 10) / d;
-					return temp.toFixed(decimals);
+					return value.toFixed(decimals);
 				}
 				return value;
 			},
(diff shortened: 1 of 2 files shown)


programmer4web:
@nomego @JockeCK @megheaiulian This is ready for review.

cosmoz-omnitable-column-number.html
```diff
@@ -93,6 +93,35 @@ <h3 style="margin: 0;">[[ title ]]</h3>
...
+				let value = item;
+				if (valuePath != null) {
+					value = this.get(valuePath, item);
+				}
+				value = this.toValue(value);
+				if (value == null) {
+					return;
+				}
+
+				const decimals = this.minimumFractionDigits;
```
nomego: This should be `maximumFractionDigits` ? (We need min/max fraction tests obviously)
Why do we need the pow/parseInt? Aren't we passed this JS float stuff with .toFixed()? :)

nomego:
Also needs rebase

cosmoz-omnitable-column-number.html
programmer4web: @nomego in the issue: https://github.com/Neovici/cosmoz-omnitable/issues/176
we have the: " 46,768 value with two decimals is 46,76

The user sets both to and from filter as 46,76.
This results that at least not the row with the value 46,768 is not shown." 

My answer is: I have tried with toFixed but it rounds up the value. We need to truncate to two decimals without rounding.
https://stackoverflow.com/questions/4187146/truncate-number-to-two-decimal-places-without-rounding

nomego: Ah, now I get it! And I wonder if the case was actually correct to begin with.
If we filter out values between 46,75 - 46,76, we shouldn't show a value with 46,768 since it's outside the range. However, if 46,768 is displayed as 46,76 then that's the error, it should be 46,77. But we should show 46,75xx.

Problem is with 46,7499 - it's outside the range but will be represented as 46,75. Same with 46,760001. I guess we should consider them 46,75 and include them, but that warrants rounding, not cutting off digits.

programmer4web:
@nomego I have implemented some tests from your latest review. Also I have used toFixed again. This is ready for review.

megheaiulian:
Needs a rebase after #195 is merged.

Issue #193, opened 2018-06-20
More tests for better coverage needed


programmer4web:
Looking at /coverage/lcov-report, basically uncovered are:
The handle of undefined, null values of parameters or properties - in multiple columns;

unrelated tests that can be done to increase coverage:
Tests for templatizer: detachInstance, releaseInstances;
cosmoz-omnitable-group-row.html: _configureTemplateInstance;
cosmoz-omnitable-item.html.html: _labelChanged;
cosmoz-omnitable-column-datetime.html: _fromHashString;

programmer4web:
[WIP]

PR #194, opened 2018-06-25
More tests for better coverage

programmer4web:
#193
Unit tests for coverage (methods handle null, undefined, test returned values..)

cosmoz-omnitable-column-autocomplete.html
cosmoz-omnitable-column-behavior.html
cosmoz-omnitable-column-datetime.html.html

### DIFF
--- a/test/basic.html
+++ b/test/basic.html
@@ -183,6 +183,24 @@
 				done();
 			});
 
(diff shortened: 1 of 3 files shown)


nomego:
@programmer4web should we merge this and open up more prs for additional tests?

PR #206, opened 2018-08-29
Aligning dropdowns within onmitable content row cells

MicoNeovici:
Jim, these are the changes Simon asked for yesterday. I´m not sure this is the best solution. Maybe you want to ask Patrik for input. It works for now though.


MicoNeovici:
Created issue instead. Closing PR

PR #208, opened 2018-08-29
Improves managing templatizers

### LINKED ISSUE #204: Disable column attribute does not update table
Changing the `disable` attribute on a cosmoz-omnitable-column does not show nor hide the column in the table when the table has already been rendered. The attribute only works when reloading the page. Related issue to this is #19628.

megheaiulian:
fixes #204

### DIFF
--- a/cosmoz-omnitable-column-behavior.html
+++ b/cosmoz-omnitable-column-behavior.html
@@ -258,22 +258,22 @@
 
 		_cellTemplateChanged: function (cellTemplate) {
 			if (cellTemplate !== undefined && cellTemplate !== null) {
-				this._dataTemplatizer = this._createTemplatizer(cellTemplate);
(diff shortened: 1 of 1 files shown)


cosmoz-omnitable-column-behavior.html
```diff
@@ -427,10 +427,10 @@
 		},
 
 		_editableChanged() {
-			if (this._dataTemplatizer) {
-				this._dataTemplatizer.releaseInstances();
+			const prevTemplatizer = this.editable ? this._cellTemplatizer : this._editableTemplatizer;
```
nomego: Wouldn't this point to editable after change? Maybe use the old/new value args instead?

megheaiulian: If `editable` has now become true we want to cleanup cellTemplate.
if `editable` has now become false we want to cleanup editableTemplate.

Issue #215, opened 2018-09-04
Updated column data does not update the table


dotpointer:
I am not sure if this is related only to `cosmoz-omnitable-column-boolean` or all columns, but updating the data sent to the table through `data` does not update the column content in the table.

Example table:
```
<cosmoz-omnitable id="values" data="[[ _dimensionValues ]]" sort-on="value">
    <cosmoz-omnitable-column
        name="value"
        title="[[ dimension.name ]]"
        value-path="value.value">
    </cosmoz-omnitable-column>
    <cosmoz-omnitable-column-boolean name="active" value-path="value.isActive"
        title="[[ _('Active', t) ]]"
        true-label="[[ _('Yes', t)]]"
        false-label="[[ _('No', t)]]"
        editable="[[ _isEditable ]]">
    </cosmoz-omnitable-column-boolean>
</cosmoz-omnitable>
```
The boolean column does not change if `_dimensionValues[...].isActive` changes in the above table.

Related RM issue for this is #19683.

megheaiulian:
@dotpointer Shouldn't `value-path` be simply `isActive`? How does the `_dimensionValues` array look like?

dotpointer:
The path to the data is `value.isActive` because the data is located in a sub object.

The full table looks like this:

```
<cosmoz-omnitable id="values" data="[[ _dimensionValues ]]" sort-on="value">
	<cosmoz-omnitable-column
		name="value"
		title="[[ dimension.name ]]"
		value-path="value.value">
	</cosmoz-omnitable-column>
	<cosmoz-omnitable-column-autocomplete
		name="description"
		title="[[ _('Description', t) ]]"
		value-path="value.description"
		editable="[[ _isEditable ]]">
	</cosmoz-omnitable-column-autocomplete>
	<cosmoz-omnitable-column-boolean name="active" value-path="value.isActive"
		title="[[ _('Active', t) ]]"
		true-label="[[ _('Yes', t)]]"
		false-label="[[ _('No', t)]]"
		editable="[[ _isEditable ]]">
	</cosmoz-omnitable-column-boolean>
	<cosmoz-omnitable-column-autocomplete disabled="[[ !dimension.useCategories ]]"
		name="category"
		title="[[ _('Category', t) ]]"
		value-path="value.category"
		editable="[[ _isEditable ]]">
		<template class="edit-cell">
			<paper-dropdown-menu-light class="categories-dropdown" label="[[ _('Select category', t) ]]">
[...]

PR #219, opened 2018-09-07
215 - Boolean column, fix to follow column data updates after table has been rendered

dotpointer:
Boolean column, fix to follow column data updates after table has been rendered, issue is #215 and #19683.

### DIFF
--- a/cosmoz-omnitable-column-boolean.html
+++ b/cosmoz-omnitable-column-boolean.html
@@ -13,8 +13,7 @@
 		<template class="edit-cell">
 			<paper-dropdown-menu>
 				<paper-listbox class="dropdown-content" slot="dropdown-content"
-					attr-for-selected="value"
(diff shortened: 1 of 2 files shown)


dotpointer:
Please note: The XLSX errors are not due to the changes in this pull request. I tried to revert the pull request to only new line and the errors were still there (reverted and force pushed over the changes after it, therefore not present in commit log). Something is broken with the XLSX export.

nomego:
Oh and however this PR turns out, it should be rebased to make tests pass and merge work.

dotpointer:
Some kind of mix I think. Changed the order of the selections and made changes so `false` is `0` and `true` is `1` and rebased.

nomego:
Can you just check to see if switching from "fallback-selection" to "selection" without changing anything else works before we merge this ?

dotpointer:
Tested and seems to work, removed the unnecessary changes.

PR #229, opened 2018-10-01
removed input type numeric. 

MicoNeovici:
Paper input and standard HTML had different opinions about what is a numeric value. Minus values would not float the label when input type was numeric.


nomego:
Ah, also type="number" has pretty big implications on mobile.

nomego:
Solved with #271

PR #239, opened 2018-12-10
20454 - Adding hide-footer attribute

dotpointer:
Adding hide-footer attribute.

Testing: Try to set the `hide-footer` on `cosmoz-omnitable`, the footer should be hidden.

Needed in #20454.

### DIFF
--- a/cosmoz-omnitable.html
+++ b/cosmoz-omnitable.html
@@ -134,7 +134,7 @@
 					</cosmoz-grouped-list>
 				</div>
 			</div>
-			<div class$="footer [[ _getFooterHiddenClass(hideFooter) ]]">
+			<div class$="footer [[ ifElse(hideFooter, 'hidden', '') ]]">
 				<div class="footer-controls">
 					<div class="footer-control">
 						<paper-dropdown-menu id="groupOnSelector" vertical-align="bottom" horizontal-align="left"
--- a/cosmoz-omnitable.html
+++ b/cosmoz-omnitable.html
@@ -134,7 +134,7 @@
 					</cosmoz-grouped-list>
 				</div>
 			</div>
-			<div class="footer">
+			<div class$="footer [[ _getFooterHiddenClass(hideFooter) ]]">
 				<div class="footer-controls">
 					<div class="footer-control">
 						<paper-dropdown-menu id="groupOnSelector" vertical-align="bottom" horizontal-align="left"
--- a/cosmoz-omnitable.html
+++ b/cosmoz-omnitable.html
@@ -14,6 +14,7 @@
 <link rel="import" href="../paper-listbox/paper-listbox.html"/>
 <link rel="import" href="../paper-spinner/paper-spinner-lite.html">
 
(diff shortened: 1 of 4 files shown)


cosmoz-omnitable.html
```diff
@@ -134,7 +134,7 @@ <h3 class="groupRow-label">
 					</cosmoz-grouped-list>
 				</div>
 			</div>
-			<div class="footer">
+			<div class$="footer [[ _getFooterHiddenClass(hideFooter) ]]">
```
nomego:
```suggestion
			<div class$="footer [[ ifElse(hideFooter, 'hidden', '') ]]">
```

cosmoz-omnitable.js
```diff
@@ -182,6 +182,13 @@
 				value: 0
 			},
 
+			/**
+			 * Footer visibility setting.
+			 */
+			hideFooter: {
+				type: Boolean
```
nomego:
```suggestion
				type: Boolean,
				value: false
```

cosmoz-omnitable.js
```diff
@@ -182,6 +182,13 @@
 				value: 0
 			},
 
+			/**
+			 * Footer visibility setting.
+			 */
+			hideFooter: {
```
nomego: Make it routine to also add test

cosmoz-omnitable.js
dotpointer: Changed to it.

cosmoz-omnitable.html
dotpointer: Changed to it, console complains: method `ifElse` not defined when testing on frontend. Maybe local problem, re-running component update.

dotpointer: Problem persists, ifElse is not defined.

dotpointer: Trying to use `cosmoz-behaviors`, since it contains `ifElse`.

nomego: Yeah sorry it seems omnitable itself doesn't use the TemplateHelperBehavior but the component depends on it. But some columns do. So yeah, please add the behavior to cosmoz-omnitable.js

dotpointer: Seems to work with it.

cosmoz-omnitable.js
dotpointer: Test is now added, thanks to @megheaiulian for guidance.

dotpointer:
I found that the bottom-bar were not shown if this attribute was used. I have made an update to the pull request. Also tried adding checks for footer-controls and footer-tableStats, but these failed in one of the Chrome browsers on the second run.

dotpointer:
@nomego The need for it is because of the demo made 2018-12-11, where a request to remove the footer was made. My intent was to hide the visible footer but not the details needed to display the bottom actions. Since this never reached the demo then this could be closed if the demo should not be put into production.

dotpointer:
Demo is over, closing this.

PR #251, opened 2019-02-23
list column - refactor data flow

nomego:
- drop unnecessary transformation through selected items vs filter
-- makes filterChanged handler unnecessary
--- makes filterChanged test unnecessary
-- improves data flow (shorter, one-way, keeps original objects)
- reuse unique array filtering logic from autocomplete
-- avoids inner loop

Prepares list to be aligned with autocomplete.

Testing made easier after merging #250 250


dotpointer:
Rebasing it, `git rebase master`.

PR #250, opened 2019-02-23
x-page/demo-behavior - Add list of objects column

nomego:
Also var -> let/const


dotpointer:
I am pushing a rebased version, just ran `git rebase master` on it, nothing else.

nomego:
I actually only tested this with #251 so maybe current implementation is broken?

Issue #246, opened 2019-02-23
Omnitable 2.x - 1/4 Columns and helpers


nomego:
A first step for 2.x would be to upgrade the columns and helpers to class-based setup since they are isolated and small and use mixinBehaviors.

nomego:
Solved by #253

PR #253, opened 2019-02-27
Class-based omnitable columns

cristinecula:
Related to #246


cosmoz-omnitable-column-list-horizontal.html
```diff
@@ -39,83 +48,100 @@
...
-				return values
-					.reduce((acc, val) => acc.concat(val), [])
-					// Make the item list unique
-					.filter((value, index, array) => {
-						if (array.indexOf(value) !== index) {
-							return false;
-						}
-						const val = this.valueProperty ? this.get(this.valueProperty, value) : value,
-							hasVal = unique.indexOf(val) !== -1;
+				return (
```
nomego: Why parenthesis?

cosmoz-omnitable-column-mixin.html
```diff
@@ -0,0 +1,545 @@
...
+				}
+
+				ready() {
+					super.ready();
+					this._headerTemplatizer = this._createTemplatizer('template.header');
+					this._itemCells = [];
+				}
+
+				_createTemplatizer(templateElementOrSelector) {
+					var template,
```
nomego: Use const where possible, otherwise let.

cosmoz-omnitable-column-mixin.html
```diff
@@ -0,0 +1,545 @@
...
+						templatizer,
+						root;
+					if (templateElementOrSelector instanceof HTMLTemplateElement) {
+						template = templateElementOrSelector;
+						templateHost = templateElementOrSelector.parentNode;
+					} else {
+						template = this.querySelector(
+							templateElementOrSelector
+						);
+						if (template) {
```
nomego:
```suggestion
						const root = this.getRootNode();
						if (!template) {
							template = this.root.querySelector(templateElementOrSelector);
						}
```

cosmoz-omnitable-column-mixin.html
```diff
@@ -0,0 +1,545 @@
...
+					}
+
+					templatizer = document.createElement('cosmoz-omnitable-templatizer');
+					this.appendChild(templatizer);
+					templatizer.init(template, templateHost);
+					return templatizer;
+				}
+
+				_cellTemplateChanged(cellTemplate) {
+					if (cellTemplate !== undefined && cellTemplate !== null) {
```
nomego: Only scenario where we use `==` is against null to capture both undefined and null:

```suggestion
					if (cellTemplate != null) {
```

cosmoz-omnitable-column-mixin.html
```diff
@@ -0,0 +1,545 @@
...
+						this.set('sortOn', valuePath);
+					}
+				}
+
+				getComparableValue(item, valuePath) {
+					return this.get(valuePath, item);
+				}
+
+				_valueChanged(event) {
+					var input = event.target,
```
nomego: const

cosmoz-omnitable-column-mixin.html
```diff
@@ -0,0 +1,545 @@
...
+						value = input.value,
+						item = event.model.item,
+						oldValue = this.get(this.valuePath, item);
+
+					this.set(this.valuePath, value, item);
+					this._fireItemChangeEvent(item, this.valuePath, oldValue);
+				}
+
+				_fireItemChangeEvent(item, valuePath, oldValue, formatFn) {
+					var value = this.get(valuePath, item),
```
nomego: const

cosmoz-omnitable-column-mixin.html
```diff
@@ -0,0 +1,545 @@
...
+				/**
+				 * Returns whether `filterNotify.base` or `filter` is set to a usable value.
+				 * @param {Object|void} filterNotify filter.*
+				 * @returns {Boolean}  True if filter or filterNotify.base is set
+				 */
+				hasFilter(filterNotify) {
+					const filter = filterNotify == null ? this.filter : filterNotify.base,
+						isArray = Array.isArray(filter);
+
+					if (
```
nomego: When if statements go on multiple lines, that's an indication they're unnecessarily complex.
isArray will return false if filter is null or '' so no need for those checks.
You can split it up into one if for arrays and one for objects to make it more readable.
```suggestion
					if (Array.isArray(filter)) {
						return filter.length > 0;
					}
					if (typeof filter === 'object')  {
						const setFilters = Object.keys(filter).filter(k => filter[k] != null);
						return setFilters.length > 0;
					}
```

cosmoz-omnitable-repeater-mixin.html
```diff
@@ -0,0 +1,279 @@
...
+					return element.__column;
+				}
+
+				_columnsChanged({ path, base, value }) {
+					let removedColumns,
+						reMatch = editableRegex.exec(path);
+
+					if (path === 'columns') {
+						if (this._elements && this._elements.length) {
+							removedColumns = this._elements.map(element => {
```
nomego: const, scoped
```suggestion
							const removedColumns = this._elements.map(element => element.__column);
```

cosmoz-omnitable-repeater-mixin.html
```diff
@@ -0,0 +1,279 @@
...
+				 * @param {Object} element - The element.
+				 * @return {Object} - The column.
+				 */
+				getElementColumn(element) {
+					return element.__column;
+				}
+
+				_columnsChanged({ path, base, value }) {
+					let removedColumns,
+						reMatch = editableRegex.exec(path);
```
nomego: const

cosmoz-omnitable-column-range-mixin.html
```diff
@@ -287,35 +325,51 @@
...
 			const limit = this._limit,
 				limitPath = path === 'min' ? 'from' : 'to',
-				limitedValue = value != null ? this.toValue(
-					this.toValue(value, this.get(limitPath + 'Min', limit), Math.max),
-					this.get(limitPath + 'Max', limit),
-					Math.min
-				) : null;
-
-			if (this.getComparableValue(value) !== this.getComparableValue(limitedValue)) {
+				limitedValue =
```
nomego: This needs to be simplified.

cristinecula: OK, but I'll do it in a separate pull request, after this one is finished.

cosmoz-omnitable-column-mixin.html
```diff
@@ -409,12 +408,18 @@
 			const filter = filterNotify == null ? this.filter : filterNotify.base,
 				isArray = Array.isArray(filter);
 
-			if (filter == null || filter === ''
-				|| isArray && filter.length === 0
-				|| typeof filter === 'object' && !isArray && Object.keys(filter).filter(k => filter[k] != null).length === 0
-			) {
+			if (!filter) {
```
nomego: Why not `== null || === ''` ?
If this is a generic filtering, for number-column a `0` filter is still a filter but would fail this condition.

cosmoz-omnitable-column-mixin.html
```diff
@@ -409,12 +408,18 @@
...
 				return false;
 			}
+
+			if (isArray && filter.length === 0) {
```
nomego: Why do we want arrays to continue the flow here?
```js
if (isArray) {
  return filter.length > 0;
}
```

Would make sure that all arrays are handled and we don't need to remember to check `isArray` later on?

cosmoz-omnitable-column-mixin.html
```diff
@@ -409,12 +408,18 @@
...
+				return false;
+			}
+
+			if (typeof filter === 'object' && !isArray && Object.keys(filter).filter(k => filter[k] != null).length === 0) {
```
nomego: Again, these conditions are too long to read and like arrays, I think we should handle all objects in the same scope to avoid having to consider them later on in the method:

```js
if (typeof filter === 'object') {
 const setFilters = Object.keys(filter).filter(k => filter[k] != null);
 return setFilters.length > 0;
}

Otherwise any added code after this will have to make sure to consider arrays, other objects and whether the object properties are set?

File: cosmoz-omnitable-item-row.html [hunk 14]
```diff
@@ -104,22 +106,17 @@
...
 
 		_itemUpdated(itemChange) {
 			if (itemChange.path === 'item') {
 				this.elements.forEach(element => {
 					this._forwardProperty(
 						this.getElementTemplateInstance(element),
-						'item',
-						itemChange.value,
-						true
+						'item',	itemChange.value,	true
```
nomego: It's fine to put the args on separate rows here, either do all on one or one per row, not mixed.

cosmoz-omnitable-column-amount.html
```diff
@@ -40,42 +40,47 @@ <h3 style="margin: 0;">[[ title ]]</h3>
...
+			[Cosmoz.TranslatableBehavior],
+			Polymer.Element
+		))) {
 			static get is() {
 				return 'cosmoz-omnitable-column-amount';
 			}
 			static get observers() {
 				return [
-					'_valuesChanged(autodetect, currency, values)'
+					'_valuesChanged(autodetect, currency, values)',
```
nomego: Extra comma

cosmoz-omnitable-column-amount.html
```diff
@@ -142,19 +143,23 @@ <h3 style="margin: 0;">[[ title ]]</h3>
...
+			 * @param {Function} limitFunc The function used to limit the number (Math.min|Math.max)
+			 * @returns {Object|void} Value converted to Number or void
+			 */
 			toAmount(value, limit, limitFunc) {
 				if (value == null || value === '') {
 					return;
 				}
 
-				if (typeof value !== 'object' || value.currency == null || value.currency === '') {
+				if (
```
nomego: one line (revert)

cosmoz-omnitable-column-amount.html
```diff
@@ -293,7 +303,9 @@ <h3 style="margin: 0;">[[ title ]]</h3>
 				if (amount == null) {
 					return null;
 				}
-				const toCurrency = this.toNumber(amount.amount) * (rates[amount.currency] || 1) / (rates[this.currency] || 1);
+				const toCurrency =
```
nomego: one line (revert)

cosmoz-omnitable-column-amount.html
```diff
@@ -304,7 +316,9 @@ <h3 style="margin: 0;">[[ title ]]</h3>
 				}
 				return this.toValue({
 					amount: number,
-					currency: property && this.get(['filter', property, 'currency']) || this.currency
+					currency:
```
nomego: one line (revert)

cosmoz-omnitable-column-boolean.html
```diff
@@ -34,63 +44,69 @@
...
-			],
+				};
+			}
 
-			observers: [
-				'_filterChanged(filter)'
-			],
+			static get observers() {
+				return [
+					'_filterChanged(filter)',
```
nomego: extra comma

cosmoz-omnitable-column-boolean.html
```diff
@@ -102,38 +118,39 @@
...
 
 			_filterChanged(filter) {
 				if (this._filterChangedFromInput) {
 					this._filterChangedFromInput = false;
 					return;
 				}
 				this._filterChangedFromAbove = true;
-				this._textFilter = typeof filter === 'boolean' ? filter.toString() : null;
-			},
+				this._textFilter =
```
nomego: one line (revert)

cosmoz-omnitable-column-datetime.html
```diff
@@ -38,78 +38,92 @@ <h3 style="margin: 0;">[[ title ]]</h3>
...
 
 			_toHashString(value) {
 				const date = this.toValue(value);
 				if (date == null) {
 					return '';
 				}
 				//Use utc in hash
-				return date.toISOString().slice(0, 19).replace(/:/g, '.');
-			},
+				return date
```
nomego: one line (revert)

cosmoz-omnitable-column-list-data.html
```diff
@@ -25,88 +25,118 @@
 		</style>
 
 		<ul hidden$="[[ isEmpty(items) ]]">
-			<li>
-				<span>[[ _firstItem(items) ]]</span>
-			</li>
+			<li><span>[[ _firstItem(items) ]]</span></li>
 			<li class="see-more" hidden$="[[_hideExpand(items, _expanded)]]">
-				<a href="#" on-tap="_toggleExpand">[[ _('and {0} more', _othersCount, t) ]]</a>
+				<a href="#" on-tap="_toggleExpand"
```
nomego: one line (revert) (or.. what happened here?)

cosmoz-omnitable-column-number.html
```diff
@@ -36,49 +36,60 @@ <h3 style="margin: 0;">[[ title ]]</h3>
...
+						type: Number,
+						value: null
+					},
+					minimumFractionDigits: {
+						type: Number,
+						value: null // browser default 0 for numbers, currency-specific or 2 for currency
+					},
+					formatter: {
+						type: Object,
+						computed:
```
nomego: one line (revert)

cosmoz-omnitable-column-range-mixin.html
```diff
@@ -113,35 +127,41 @@
...
 				return '';
 			}
 			return value;
-		},
+		}
 
 		getString(item, valuePath = this.valuePath) {
 			if (valuePath === undefined) {
-				console.error(this, 'has undefined valuePath', valuePath, 'for item', item);
+				console.error(
```
nomego: one line (revert)

cosmoz-omnitable-column-range-mixin.html
```diff
@@ -276,7 +312,9 @@
 		_limitInput() {
 			const input = this._filterInput,
 				path = this.__inputChangePath,
-				value = path ? this._fromInputString(this.get(path, input), path) : null;
+				value = path
```
nomego: one line (revert)

cosmoz-omnitable-column-range-mixin.html
```diff
@@ -287,35 +325,45 @@
...
 
 		_updateFilter() {
 			const input = this._filterInput,
 				filter = this.filter,
 				min = this._fromInputString(input.min, 'min'),
 				max = this._fromInputString(input.max, 'max');
 
-			if (this.getComparableValue(min) === this.getComparableValue(filter, 'min')
-				&& this.getComparableValue(max) === this.getComparableValue(filter, 'max')
+			if (
```
nomego: two lines (revert)

cosmoz-omnitable-column-range-mixin.html
```diff
@@ -326,8 +374,11 @@
...
```
nomego: two lines (revert)

cosmoz-omnitable-column-range-mixin.html
```diff
@@ -336,17 +387,20 @@
...
 			});
-			this.cancelDebouncer('_limitInput');
-		},
+			this._limitInputDebouncer.cancel();
+		}
 
 		hasFilter() {
 			const filter = this.filter,
-				isSet = Cosmoz.OmnitableColumnBehavior.hasFilter.apply(this, arguments);
+				isSet = super.hasFilter.apply(
```
nomego: one line

cosmoz-omnitable-column-time.html
```diff
@@ -83,34 +95,40 @@ <h3 style="margin: 0;">[[ title ]]</h3>
...
 
 			_toHashString(value) {
 				const date = this.toValue(value);
 				if (date == null) {
 					return '';
 				}
 				//Use utc in hash
-				return date.toISOString().slice(11, 19).replace(/:/g, '.');
-			},
+				return date
```
nomego: one line (revert)

cosmoz-omnitable-repeater-mixin.html
```diff
@@ -0,0 +1,280 @@
...
+
+				_forwardProperty(instance, name, value, flush = false) {
+					instance._setPendingProperty(name, value);
+
+					if (flush) {
+						instance._flushProperties(true);
+					}
+				}
+
+				_forwardNotifyPath(
```
nomego: one line for declarations

cristinecula:
Thanks for the review @nomego!

cosmoz-omnitable-column-amount.html
cristinecula: Should I add a rule to eslint for this?

https://eslint.org/docs/rules/comma-dangle

Something along the lines of:
```json
"comma-dangle": [
	"error",
	{
		"arrays": "never",
		"objects": "only-multiline",
		"imports": "never",
		"exports": "never",
		"functions": "ignore"
	}
],
```

cosmoz-omnitable-column-mixin.html
cristinecula: Of course, you are right! I'll be more careful about changing the meaning of the code.

cosmoz-omnitable-column-mixin.html
cristinecula: Awesome, thanks!

cristinecula:
I will revert all unnecessary changes that I've added due to dev environment misconfiguration (I had prettier running).

cosmoz-omnitable-column-amount.html
nomego: Actually I was under the impression that dangling commas could cause issues outside of IE8, but if that's not the case maybe we should rather enable/enforce it for multi-line arrays/objects (and also prevent it in all other cases).

cosmoz-omnitable-column-boolean.html
```diff
@@ -1,31 +1,41 @@
...
 			<paper-dropdown-menu>
-				<paper-listbox class="dropdown-content" slot="dropdown-content"
+				<paper-listbox
+					class="dropdown-content"
+					slot="dropdown-content"
 					attr-for-selected="value"
 					selected="[[ _computeSelected(item, valuePath) ]]"
-					on-selected-item-changed="_valueChanged">
+					on-selected-item-changed="_valueChanged"
+				>
```
cristinecula: @nomego is there a style guide I can consult for formatting templates?
If not, should I revert these changes? I think that when formatting elements with many attributes, we could follow the rule that's used for function arguments: "all arguments on one line if it fits, each argument on a new line if it doesn't", to keep things consistent. 
Also, having the closing bracket on a separate line helps with diffing when new attributes are added (just like comma-dangle).

nomego: Sure use those guidelines but don't update the templates just to fix it at this point.
But whenever you create/edit a template you can use that guideline.
I suspect things might change somewhat when we (hopefully in a near future) start moving to template getters.

cristinecula: Alright! In this case I will revert the changes to the templates.

cosmoz-omnitable-column-amount.html
cristinecula: This will affect many files, so I will do it in another pull request. #254

cristinecula:
Done. The pull request should be much cleaner now.

cosmoz-omnitable-column-range-mixin.html
```diff
@@ -287,20 +307,18 @@
 
 			const limit = this._limit,
 				limitPath = path === 'min' ? 'from' : 'to',
-				limitedValue = value != null ? this.toValue(
```
nomego: This declaration avoids a lot of code/evaluations if value is null, which we don't now. What happens when we call toValue() with a null/undefined value?

cosmoz-omnitable-group-row.html
```diff
@@ -1,109 +1,123 @@
...
+					instance._flushProperties(true);
+				}
+			}
 
-		_columnChanged(newColumn) {
-			if (!newColumn) {
-				return;
+			_itemUpdated(itemChange) {
+				if (itemChange.path === 'item') {
+					this.elements.forEach(function (element) {
```
nomego: Should use arrow function and avoid binding "this".

cosmoz-omnitable-column-range-mixin.html
cristinecula: There is already a null value test on a previous line which returns early, preventing unnecessary evaluations: https://github.com/Neovici/cosmoz-omnitable/blob/3562c54c4bc3228ddd7195e0c24fb0ebb6296274/cosmoz-omnitable-column-range-mixin.html#L303-L312

nomego: Ah right, missed that!
Nice work on the readability though, I think we could also break out
`minValue = this.toValue(value, lowerLimit, Math.max),`

Issue #254, opened 2019-03-05
Add eslint comma-dangle rule


cristinecula:
Actually I was under the impression that dangling commas could cause issues outside of IE8, but if that's not the case maybe we should rather enable/enforce it for multi-line arrays/objects (and also prevent it in all other cases).

_Originally posted by @nomego in https://github.com/Neovici/cosmoz-omnitable/pull/253_

```json
"comma-dangle": [
	"error",
	{
		"arrays": "always-multiline",
		"objects": "always-multiline",
		"imports": "never",
		"exports": "never",
		"functions": "ignore"
	}
],
```

nomego:
This has been decided against, any further eslint rule changes should be against https://github.com/Neovici/eslint-config

PR #257, opened 2019-03-06
(patch #254) Add eslint comma-dangle rule

cristinecula:
I have added the rule and run --fix on the whole repository.


cristinecula:
The main reason for using dangling commas is improved diff/blame experience. Adding or removing an element to an array or object will display in the diff as only affecting those lines. Without dangling commas, when doing `git blame`, you will often find lines that were only changed by adding/removing a comma, thus making it harder to identify the relevant commit.

PR #258, opened 2019-03-06
Upgrade <cosmoz-omnitable> to Polymer 2.x

cristinecula:
* Upgrade to class-based element
* remove use of Polymer.dom
* use Polymer.Debouncer instead of this.debounce
* use addEventListener/removeEventListener instead of listeners
* use FlattenedNodesObserver instead of Polymer.dom().observe

@megheaiulian @nomego

### DIFF
--- a/cosmoz-omnitable-repeater-mixin.html
+++ b/cosmoz-omnitable-repeater-mixin.html
@@ -163,7 +163,7 @@
 
 		_addElements(start, count) {
 			const end = start + count,
-				parent = Polymer.dom(this);
+				parent = this;
 
 			for (let i = start; i < end; i++) {
 				const column = this.columns[i],
@@ -182,7 +182,7 @@
 				this._configureTemplateInstance(instance);
 				element.setAttribute('slot', this._slotName);
 
-				Polymer.dom(element).appendChild(instance.root);
+				element.appendChild(instance.root);
 
 				if (i < this._elements.length) {
 					parent.insertBefore(element, this._elements[i]);
@@ -207,7 +207,7 @@
 				.forEach(element => {
 					this._detachTemplateInstance(element.__instance, element.__column, element);
 					element.__instance = element.__column = element.column = null;
-					Polymer.dom(this).removeChild(element);
+					this.removeChild(element);
 				});
 		}
 
--- a/cosmoz-omnitable-repeater-mixin.html
+++ b/cosmoz-omnitable-repeater-mixin.html
@@ -162,8 +162,7 @@
 		}
 
 		_addElements(start, count) {
-			const end = start + count,
-				parent = this;
+			const end = start + count;
 
 			for (let i = start; i < end; i++) {
 				const column = this.columns[i],
@@ -185,9 +184,9 @@
 				element.appendChild(instance.root);
 
 				if (i < this._elements.length) {
-					parent.insertBefore(element, this._elements[i]);
+					this.insertBefore(element, this._elements[i]);
 				} else {
-					parent.appendChild(element);
+					this.appendChild(element);
 				}
 
 				// HACK(plequang): repeated element should have style scoped to cosmoz-omnitable, not to the column element.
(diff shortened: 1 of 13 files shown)


cristinecula:
I have profiled the code before (d52c5670) and after the upgrade (1faccc48) and there is a significant improvement in initialization and resize time.

You can load the [profiles](https://filebin.ca/4ZFWjBW92v5g) in chrome dev tools > performance tab.

# Initialization improvements

This is the profile for page initialization. There are 11 idle callbacks needed to render the whole iron-list, after the initial rendering. This is caused by the total time needed to render a row (~50ms). At most two row renders fit in an idle callback.
image

After the upgrade, only 5 idle callbacks are necessary to render the whole iron-list, due to the reduced render time of a row. In 1faccc48 you can see one of them, but there are more performance improvements possible (for example: item-expand should render the columns only when expanded).
image

# resize improvements

Before the upgrade there were significant lags when resizing the table, caused by adding and removing list columns:
image

After the upgrade:
image

There are still some blocks, but further investigation is required.

cosmoz-omnitable-repeater-mixin.html
```diff
@@ -163,7 +163,7 @@
 
 		_addElements(start, count) {
 			const end = start + count,
-				parent = Polymer.dom(this);
+				parent = this;
```
nomego: We should just drop the `parent` const completely and use `this` rather.

cosmoz-omnitable.js
```diff
@@ -275,65 +311,109 @@
...
+				this._groupItemsDebouncer.flush();
+			}
+
+			if (this._sortItemsDebouncer) {
+				this._sortItemsDebouncer.flush();
+			}
+		}
+
+		_cancelDebouncers() {
+			[this._adjustColumnsDebouncer,
```
nomego: Not sure how to make this look nice and readable.. 

```suggestion
			[
				this._adjustColumnsDebouncer,
				this._updateColumnsDebouncer,
				this._filterItemsDebouncer,
				this._groupItemsDebouncer,
				this._sortItemsDebouncer
			]
				.filter(d => d != null)
				.forEach(d => d.cancel());
```

maybe?

`d != null` is both more expressive and faster (truthy/falsy statements is about 8 checks vs `!= null` at about 2)

cosmoz-omnitable.js
```diff
@@ -992,7 +1087,7 @@
 		 */
 		_applySortingDirection(e) {
 			const column = e.model.column,
-				data = Polymer.dom(e).localTarget.dataset,
+				data = e.target.dataset,
```
nomego: Why target instead of localTarget?

cosmoz-omnitable.js
```diff
@@ -1234,5 +1329,12 @@
 					button.$.dropdown.fitInto = this;
 				});
 		}
-	});
+
+		_debounce(name, fn, asyncModule = Async.timeOut.after(0)) {
+			this[name] = Debouncer.debounce(this[name], asyncModule, fn);
```
nomego: Might want to have a debouncer namespace (`this.debouncers[key]`)

nomego:
Btw profiling looks great! Can't wait to try this out on alpha!

cosmoz-omnitable.js
cristinecula: `localTarget` is a property that is added by Polymer.dom to the event; the native equivalent is event.target.

The upgrade guide also mentions this:

> Class-based elements should use the native DOM APIs instead of the Polymer.dom APIs. Legacy elements can optionally use the native APIs.
> 
> For standard DOM operations, remove the Polymer.dom() wrapper.
> Use this.shadowRoot in place of Polymer.dom(this.root).
> For events, use the standard v1 event API:
> - Polymer.dom(event).localTarget becomes event.target.
> - Polymer.dom(event).path becomes event.composedPath().
> - Polymer.dom(event).rootTarget becomes event.composedPath()[0].

https://polymer-library.polymer-project.org/2.0/docs/upgrade

cosmoz-omnitable.js
cristinecula: This also simplifies the `cancelDebouncers` method

cristinecula:
Finished incorporating the review feedback. I also upgraded the treenode-column, but the tests will fail until this is released.
https://github.com/Neovici/cosmoz-omnitable-treenode-column/pull/32

PR #264, opened 2019-03-08
(patch) Update Polymer 2.x imports

### LINKED ISSUE #263: Update Polymer 2.x imports
Something that we missed in #258 was to update the html imports from `polymer.html` to `polymer-element.html`.

Will hopefully drop all legacy helpers (but might also highlight hidden dependencies on them).
As soon as this is merged I'll release v2.0.0

cristinecula:
I have updated each file with only the required imports.

Fixes #263


cristinecula:
@megheaiulian

nomego:
Great! Why do we need the legacy/class import?

cristinecula:
For mixinBehaviors and for Polymer.Templatizer.

cristinecula:
I'm not sure why the Travis build failed... I have run the tests locally and it's all good.

PR #267, opened 2019-03-10
Feature/neov 295 chrome 74 tests

megheaiulian:
Investigating chrome 74 tests...

### DIFF
--- a/cosmoz-omnitable-column-date-mixin.html
+++ b/cosmoz-omnitable-column-date-mixin.html
@@ -73,17 +73,31 @@ Cosmoz.DateColumnMixin = Polymer.dedupingMixin(base =>
 			if (value == null || value === '') {
 				return;
 			}
-			const date = value instanceof Date ? value : new Date(value);
+
+			// convert value to Date
+			let date = value;
+			if (!(date instanceof Date)) {
+				// if the value is an ISO string, make sure that it has an explicit timezone
+				if (typeof value === 'string') {
+					date = this.getAbsoluteISOString(date);
+				}
+
+				date = new Date(date);
+			}
+
 			if (Number.isNaN(date.getTime())) {
 				return null;
 			}
+
 			if (limitFunc == null || limit == null) {
 				return date;
 			}
-			let lDate = this.toDate(limit);
+
+			const lDate = this.toDate(limit);
 			if (lDate == null) {
 				return date;
 			}
+
 			const comparableDate = this.getComparableValue(date),
 				comparableLDate = this.getComparableValue(lDate),
 				limitedValue = limitFunc(comparableDate, comparableLDate);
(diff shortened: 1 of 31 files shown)


cosmoz-omnitable-column-date-mixin.html
```diff
@@ -73,17 +73,31 @@
...
+
+			// convert value to Date
+			let date = value;
+			if (!(date instanceof Date)) {
+				// if the value is an ISO string, make sure that it has an explicit timezone
+				if (typeof value === 'string') {
+					date = this.getAbsoluteISOString(date);
+				}
+
+				date = new Date(date);
```
nomego: Is this if date would be a number?

cosmoz-omnitable-column-date-mixin.html
```diff
@@ -73,17 +73,31 @@
 			if (value == null || value === '') {
 				return;
 			}
-			const date = value instanceof Date ? value : new Date(value);
+
+			// convert value to Date
+			let date = value;
+			if (!(date instanceof Date)) {
+				// if the value is an ISO string, make sure that it has an explicit timezone
+				if (typeof value === 'string') {
```
nomego: Please add some comment about Safari and timezone handling here, especially if it becomes uniform so we can remove the workaround.

cosmoz-omnitable-column-time.html
```diff
@@ -142,7 +142,8 @@ <h3 style="margin: 0;">[[ title ]]</h3>
 				if (value == null) {
 					return;
 				}
-				value = this.toValue(this._fixedDate + 'T' + value);
+				value = this.getAbsoluteISOString(this._fixedDate + 'T' + value);
```
nomego: no need to put on two lines?
```suggestion
				value = this.toValue(this.getAbsoluteISOString(this._fixedDate + 'T' + value));
```

demo/ajax-demo.html
```diff
@@ -74,26 +74,25 @@ <h3>Cosmoz omnitable demo</h3>
 	</template>
```
nomego: Reformatting this file shouldn't be part of the PR, and if you touch on code like this, make sure to ES6-update it.

demo/custom-template-demo.html
```diff
@@ -87,65 +87,63 @@ <h3>Cosmoz omnitable templates demo</h3>
 	</x-page>
 
 	<script type="text/javascript">
-/*global document, window, Cosmoz, Polymer */
```
nomego: Same as ajax-demo

test/range-date.html
```diff
@@ -161,14 +160,14 @@
 					max: '2023-03-21'
 				};
 				Polymer.flush();
-				assert.equal(column.filter.min.getTime(), new Date('2015-03-18T00:00:00').getTime());
```
nomego: Why are these tests updated?

cosmoz-omnitable-column-date-mixin.html
megheaiulian: Added comments and TODO.

test/range-date.html
megheaiulian: Updated because we don't want to actually test date parsing in the assert expectation.

demo/ajax-demo.html
megheaiulian: We'll do it in another PR.

PR #269, opened 2019-03-11
Rework templatizing

### LINKED ISSUE #249: Omnitable 2.x - 3/4 Rework templatizing
Polymer 2.x, the new templatizing and the class-based approach offers new opportunities.
Can iron-list simply be extended to include the grouping support currently handled by grouped-list to avoid that middle layer or at least slim it down?
Will there be any changes to omnitable template handling (cell/edit/header)?

Should be implemented after #246, #248 and any column setup changes in regards to templatizing handled as part of this step.

cristinecula:
Fixes #249


cristinecula:
@nomego @megheaiulian Along with extracting the TemplatizeMixin, I have refactored the RepeaterMixin and simplified the repeater rows a great deal. Can you do a quick review please?

cosmoz-omnitable-templatize-mixin.html
```diff
@@ -0,0 +1,189 @@
+<script>
+(() => {
+	'use strict';
+
+	/**
+	 * Prepares instances of templates
+	 * @polymer
+	 * @mixinFunction
+	 */
+	Cosmoz.OmnitableTemplatizeMixin = Polymer.dedupingMixin(base => class extends base {
```
megheaiulian: @cristinecula There's one more thing we should handle here. 
We need to do a memory profile and check if `disconnectedCallback` the instances are properly released from memory.
If they are than we don't need to do anything.

cristinecula: I have added a cleanup callback in c8664a104896d7ee271005a4f9c4ba3c2c44d4e7

cristinecula:
This should be it. I'm updating the documentation and it should be good for review.

cristinecula:
@nomego I've added some integration tests in `test/templatize.html` that ensure that all of the features of TemplatizeMixin and RepeaterMixin are working OK. If you have suggestions for extra tests, I'll be happy to add them.

cristinecula:
In this update I have removed the `cell-template` property and functionality from OmnitableColumnMixin, thinking that it was not used anymore. It seems that this property was used to set a custom template at runtime.

This is some sample code from `full-demo.html`:

```html
<cosmoz-omnitable-column title="Custom template"
	name="custom-name" value-path="name" sort-on="name" width="50px" flex="2"
	cell-template="[[ customNameTemplate ]]">
</cosmoz-omnitable-column>
```

```js
// in x-page element
attached() {
	/**/
	Polymer.dom(this).querySelector('#customNameTemplate');
},
```

```html
<x-page class="fit">
	<template id="customNameTemplate">
		<span style="background: red;" on-tap="onTap">[[ item.name ]]</span>
	</template>
</x-page>
```

This is the equivalent working code:

```html
<cosmoz-omnitable-column title="Custom template"
	name="custom-name" value-path="name" sort-on="name" width="50px" flex="2">
		<template class="cell">
			<span style="background: red;" on-tap="onTap">[[ item.name ]]</span>
		</template>
</cosmoz-omnitable-column>
```

@nomego Do you think it's OK for me to remove this functionality? Or should I add it back, plus tests for it?

cristinecula:
Travis tests will pass after #267 is merged.

nomego:
Well I'm not sure we have a use case to inject a new custom template after init.
I think we can merge this without that functionality and add it in later if needed.

megheaiulian:
This should be merged only after #267 is merged.

cristinecula:
I've added unit tests for OmnitableTemplatizeMixin and OmnitableRepeaterMixin, and improved the semantics a bit. 
This brings the coverage up to:
```
cosmoz-omnitable-repeater-mixin.html   94.12%
cosmoz-omnitable-templatize-mixin.html 98.18%
```
As for the other components, I will increase their test coverage when I work on issues relating to them, so we can merge this.

PR #271, opened 2019-03-17
Check if label should float based on validity

nomego:
Number inputs can have allowed characters that aren't numbers (-,e) and won't
trigger a value change and thus not float the label.

However, the validity will report badInput so we can trigger a label float by
setting it to something truthy but still not visible.

Fixed in paper-input 3.x

See #229
RM Solves #19930


nomego:
Ok this is a bit crazy, it seems impossible to properly emulate a keypress (that actually fills the native input with something).
These tests doesn't actually test the behavior at all
It would have been enough just to fire an Input-event (it would have tested as much) to trigger the handler.

cosmoz-omnitable-column-number.html
```diff
@@ -6,6 +6,7 @@
 
 <link rel="import" href="../paper-input/paper-input.html">
 <link rel="import" href="../paper-dropdown-menu/paper-dropdown-menu.html">
+<link rel="import" href="../neon-animation/web-animations.html">
```
Wurper: Just a missing import or does it have anything to do with the PR?

nomego: When running the new test and clicking paper-dropdown-menu, KeyframeEffect or something like that wasn't defined.
I guess since you don't have to use animations from paper-dropdown-menu it isn't a required dependency and normally we have web-animations loaded when the animation is supposed to start?
Anyway I found a github issue for it and they solved it by adding that import. :man_shrugging:

cristinecula:
@nomego You could try using [`MockInteractions.keyEventOn`](https://github.com/Polymer/tools/blob/master/packages/modulizer/fixtures/packages/paper-button/source/bower_components/iron-test-helpers/mock-interactions.js#L397). I'm not sure if it does anything different from what you already are doing.

nomego:
Well I tried pressAndReleaseKeyOn but it does absolutely nothing in regards to interacting with the input.

cristinecula:
Yeah, I guess this interaction can only be tested via selenium automation or https://www.cypress.io/.

PR #279, opened 2019-05-02
saving csv created faulty content with horizontal list columns

JockeCK:
Checks for set textProperty in horizontal list column. If so, uses that value instead for csv string. 

Possible test candidate is order 22679_30_7365563740132_F , saving orderrows as csv. should be the text values instead of "Object Object" .

refs RM #21223

*edit*
Did the same for "save as xlsx".

### DIFF
--- a/cosmoz-omnitable-column-list-horizontal.html
+++ b/cosmoz-omnitable-column-list-horizontal.html
@@ -158,6 +158,18 @@
 
 				return value.join(', ');
 			}
+
+			toXlsxValue(item, valuePath = this.valuePath) {
+				const textProp = this.textProperty,
+					value = super.getString.apply(this, arguments);
+				if (!valuePath || !Array.isArray(value) || value.length === 0) {
+					return '';
+				}
+				if (textProp != null) {
+					return value.map(val => this.get(textProp, val)).filter(Boolean).join(', ');
+				}
+				return this.getString(item, valuePath);
+			}
 		}
 
 		customElements.define(
--- a/cosmoz-omnitable-column-list-horizontal.html
+++ b/cosmoz-omnitable-column-list-horizontal.html
@@ -56,7 +56,7 @@
 		 * @appliesMixin Cosmoz.OmnitableColumnMixin
 		 * @appliesMixin Cosmoz.ListColumnMixin
 		 */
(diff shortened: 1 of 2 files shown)


cosmoz-omnitable-column-list-horizontal.html
```diff
@@ -146,13 +146,30 @@
 			}
 
 			getString() {
```
nomego: Drop `getString()` to make it use https://github.com/Neovici/cosmoz-omnitable/blob/master/cosmoz-omnitable-column-list-mixin.html#L26
Possibly add null-check filter in that method.

cosmoz-omnitable-column-list-horizontal.html
```diff
@@ -146,13 +146,30 @@
...
 				}
 
+				if (textProp !== null) {
+					return value.map(val => this.get(textProp, val)).filter(Boolean).join(', ');
+				}
+
 				return value.join(', ');
 			}
+
+			toXlsxValue(item, valuePath = this.valuePath) {
```
nomego:
```suggestion
			toXlsxValue(item, valuePath = this.valuePath) {
			  return this.getString(item, valuePath);
			}
```

cosmoz-omnitable-column-list-horizontal.html
JockeCK: Turns out it is the getString in cosmoz-omnitable-column-mixin.html
that gets called.  How do we want to handle this? should we go back to implement an version inside list-horizontal?

nomego: Ah! Seems like the mixins are in the wrong order:

https://github.com/Neovici/cosmoz-omnitable/blob/master/cosmoz-omnitable-column-list-horizontal.html#L59

We should make sure `OmnitableColumnListHorizontal` extends `ListColumnMixin` which in turn should mixin `OmnitableColumnMixin`

JockeCK: Yeah, that worked great. Made the changes and added nullcheck. Did a quick test and it seems to work for both csv and excel

cosmoz-omnitable-column-list-horizontal.html
JockeCK: did so

PR #300, opened 2019-08-20
(patch) Range column behavior update (amount, number)

### LINKED ISSUE #299: Range column behavior update (amount, number)
As outlined in https://github.com/Neovici/cosmoz-omnitable/issues/60#issuecomment-521186490:

  - From: Value/query should update on blur
  - From: Enter, tab will move to "To" field
  - To: Value/query should update on blur, menu should close
  - To: Enter will validate input in max/min range
    - If outside range, update with valid value, perform query
    - If inside range, (blur to perform query), close menu
  - From/To: Add clear icon
  - Menu - Focus in "From" when opened

https://github.com/Neovici/cosmoz-omnitable/issues/110 is also relevant but nothing that needs to be directly addressed in this PR.

cristinecula:
Fixes #299

### DIFF
--- a/cosmoz-omnitable-column-amount.js
+++ b/cosmoz-omnitable-column-amount.js
@@ -34,20 +34,34 @@ class OmnitableColumnAmount extends rangeColumnMixin(columnMixin(translatable(
 		</template>
 
 		<template class="header" strip-whitespace>
+			<style>
+			paper-input > div > iron-icon {
+				display: none;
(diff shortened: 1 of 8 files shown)


cristinecula:
Work in progress.

cristinecula:
@nomego the implementation is done, but I still need to add tests for the new functionality.

cristinecula:
@nomego ready for review!

PR #303, opened 2019-08-29
Changes when filtering is done in an default textinput column

JockeCK:
Now filtering starts when user presses enter, blur occurs or 1 second passes without additional user text input.

Unsure if this can make it to the circuit release since alpha has started. Leaving project blank

Refs https://github.com/Neovici/cosmoz-omnitable/issues/60

### DIFF
--- a/cosmoz-omnitable-column.js
+++ b/cosmoz-omnitable-column.js
@@ -59,6 +59,11 @@ class OmnitableColumn extends columnMixin(PolymerElement) {
 					return this._getDefaultFilter();
 				},
 				observer: '_inputValueValueChanged'
+			},
+
+			_timeForNoInput: {
+				type: Number,
+				value: 1000
 			}
 		};
 	}
@@ -95,7 +100,7 @@ class OmnitableColumn extends columnMixin(PolymerElement) {
 			return;
 		}
 		this._debouncer = Debouncer.debounce(this._debouncer,
-			timeOut.after(1000),
+			timeOut.after(this._timeForNoInput),
 			() => {
 				this.filter = this.inputValue;
 			});
--- a/cosmoz-omnitable-column.js
+++ b/cosmoz-omnitable-column.js
@@ -7,6 +7,9 @@ import './ui-helpers/cosmoz-clear-button';
 import { columnMixin } from './cosmoz-omnitable-column-mixin';
 import { PolymerElement } from '@polymer/polymer/polymer-element';
 import { html } from '@polymer/polymer/lib/utils/html-tag';
+import { timeOut } from '@polymer/polymer/lib/utils/async.js';
+import { Debouncer } from '@polymer/polymer/lib/utils/debounce.js';
+import { enqueueDebouncer } from '@polymer/polymer/lib/utils/flush.js';
 
 /**
  * @polymer
@@ -47,6 +50,15 @@ class OmnitableColumn extends columnMixin(PolymerElement) {
 			editMinWidth: {
 				type: String,
 				value: '55px'
+			},
+
+			inputValue: {
+				type: Object,
+				notify: true,
+				value() {
+					return this._getDefaultFilter();
+				},
+				observer: '_inputValueValueChanged'
 			}
 		};
 	}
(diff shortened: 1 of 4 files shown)


cosmoz-omnitable-column.js
```diff
@@ -61,6 +61,7 @@ class OmnitableColumn extends columnMixin(PolymerElement) {
 
 	resetFilter() {
 		this.filter = null;
+		this.filterCandidate = null;
```
Wurper:
```suggestion
		this.inputValue = null;
```

JockeCK:
Seems I am failing 15 tests or so. will take an look

cosmoz-omnitable-column.js
```diff
@@ -61,6 +73,33 @@ class OmnitableColumn extends columnMixin(PolymerElement) {
...
+		}
+	}
+
+	_inputValueValueChanged() {
+		if (this.inputValue === '') {
+			this.filter = this.inputValue;
+			return;
+		}
+		this._debouncer = Debouncer.debounce(this._debouncer,
+			timeOut.after(1000),
```
nomego: Do we want this as a configurable property?
Or maybe at least a constant in the file?

cosmoz-omnitable-column.js
```diff
@@ -61,6 +73,33 @@ class OmnitableColumn extends columnMixin(PolymerElement) {
...
+		case 13: // Enter
+			event.preventDefault();
+			this.filter = this.inputValue;
+			break;
+		}
+	}
+
+	_inputValueValueChanged() {
+		if (this.inputValue === '') {
+			this.filter = this.inputValue;
```
nomego: Maybe add a test step and assert for this?

cosmoz-omnitable-column.js
JockeCK: See latest commit: Like this?

cosmoz-omnitable-column.js
JockeCK: See latest commit: Like this?

JockeCK:
@cristinecula I am getting an failed test in column-number.html:
out of range values are limited on blur ‣
Error: expected 2 to equal 5
  Context.<anonymous> at column-number.html:144

It is the flush(I think) I do when I added two new tests in basic.html. 

Any idea what the best way is to do here?

test/basic.html
```diff
@@ -119,6 +122,27 @@
 				assert.equal(column.sortOn, 'sortOnValuePath');
 			});
 
+			test('changing inputValue updates filter eventually', () => {
+				const column = omnitable.columns.find(col => {
```
nomego:
```suggestion
				const column = omnitable.columns.find(col => col.name === 'columnWithFilter');
```

test/basic.html
```diff
@@ -58,6 +60,7 @@
 	<script type="module">
 		import { setupOmnitableFixture } from './helpers/utils';
 		import { generateTableDemoData } from '../demo/table-demo-helper';
+		import { flush as flush$0 } from '@polymer/polymer/lib/utils/flush';
```
nomego: Do we need to rename flush? Why ?

cristinecula:
@JockeCK When working on these tests, I encountered a weird issue regarding blur events, which may cause your failing test: chrome ignores blur/focus events when the browser window is not focused. This means that if you run the tests interactively, you must always keep the window focused, otherwise you will have intermittent test failures. This is not an issue when running the tests in headless browsers. Can you confirm that this is not the cause of the test failure?

JockeCK:
> @JockeCK When working on these tests, I encountered a weird issue regarding blur events, which may cause your failing test: chrome ignores blur/focus events when the browser window is not focused. This means that if you run the tests interactively, you must always keep the window focused, otherwise you will have intermittent test failures. This is not an issue when running the tests in headless browsers. Can you confirm that this is not the cause of the test failure?

@cristinecula You are right! Kept the window up and focused and the test did not fail. Thank you!

cristinecula:
@JockeCK great! I think you should add a paragraph to `test/index.html` to mention this fact.

test/basic.html
JockeCK: Refactored, and all the rest in basic.html

test/basic.html
JockeCK: No need, removed renaming.

test/index.html
```diff
@@ -8,6 +8,8 @@
 	<script src="/components/wct-mocha/wct-mocha.js"></script>
 </head>
 <body>
+	<!-- If running tests interactively, some tests may fail if the window is not focused during the tests.
```
cristinecula: @JockeCK Maybe it's a good idea to make this text visible, so it's seen when tests are run interactively.

JockeCK: yes

PR #351, opened 2020-03-25
feat: allow objects as autocomplete values

nomego:
Remaps `values` from object to array if not passed an array.


nomego:
Switched to Object.entries(). Still good?

PR #353, opened 2020-04-06
feat: list column value property groupon

dotpointer:
Testing, check that the group on for the list column groups as it should.

Note, this does not cover the design error for the and ... more links when using cell templates.

If you know how to solve that please tell me.

Ran tests locally and they passed.

Issue is RM:23196.

### DIFF
--- a/cosmoz-omnitable-column-list.js
+++ b/cosmoz-omnitable-column-list.js
@@ -143,5 +143,24 @@ class OmnitableColumnList extends	listColumnMixin(columnMixin(translatable(
 	_getDefaultFilter() {
 		return [];
 	}
+
+	/**
+	 * Get the comparable value of an item.
+	 *
+	 * @param {Object} item Item to be processed
+	 * @param {String} valuePath Property path
+	 * @returns {String|void} Valid value or void
+	 */
+	getComparableValue(item, valuePath) {
+		const value = super.getComparableValue(item, valuePath);
+		if (this.valueProperty == null) {
+			return value;
+		}
+		const subValues = value.reduce((acc, subItem) => {
+			acc.push(this.get(this.valueProperty, subItem));
+			return acc;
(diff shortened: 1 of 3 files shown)


cristinecula:
@dotpointer how does this design error look like?

Also, should we sort the list items when grouping?

Otherwise:
user1, user2
won't group with
user2, user1

@nomego *should* they be grouped together?

dotpointer:
> @dotpointer how does this design error look like?

bild

> Also, should we sort the list items when grouping?

Yes we probably should. In this PR I take all value-property values and i concatenate them with a space. Maybe there is a safer way to do this.

cristinecula:
> bild

Well... I think we should change cosmoz-omnitable-gorup-row to forward a "grouped: true" property to the cell template and pass that to cz-link-list, either as a class or as a property. If passed as a class (`.is-grouped` for example), you can use the `:host(.is-grouped)` selector to transform the ul to an inline list.

> Yes we probably should. In this PR I take all value-property values and i concatenate them with a space. Maybe there is a safer way to do this.

I think you could just `sort` before `join`.

dotpointer:
> I think you could just `sort` before `join`.

Added this for a start.

dotpointer:
I have pushed a fix for the design problem, and also made a PR for frontend since it involves changes there too.

But this does not work in mobile mode, it still flows outside there.

dotpointer:
Fixed the overflow problem.

New problem on mobile is that the checkbox and the arrow to the right gets centered when there are multiple lines. Mostly happens on mobile mode.

cristinecula:
LGTM, considering it's such an edge case (group by list on mobile).

cristinecula:
I want to take this for a spin first, but don't have time today. If anyone else from @Neovici/frontend wants to review this, I'm OK with it :)

PR #354, opened 2020-04-09
feat: list column get comparable value test

dotpointer:
Adding test for list column getComparableValue function since this is preventing a new release from being made.

### DIFF
--- a/test/list.test.js
+++ b/test/list.test.js
@@ -117,5 +117,6 @@ suite('horizontal', () => {
 	test('returns the value of an item', () => {
 		column.valueProperty='id';
 		assert.strictEqual(column.getComparableValue({ id: 13 }, 'id'), 13);
+		assert.isUndefined(column.getComparableValue({}, 'nonexisting-path'));
 	});
 });
--- a/test/list.test.js
+++ b/test/list.test.js
@@ -113,4 +113,9 @@ suite('horizontal', () => {
 		column.textProperty = null;
 		assert.equal(column.getString({ list: ['123', '345', '678']}), '123, 345, 678');
 	});
+
+	test('returns the value of an item', () => {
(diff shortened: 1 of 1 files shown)


cristinecula:
@dotpointer can you fix the lint error?

dotpointer:
> @dotpointer can you fix the lint error?

Fixed it now and pushed it.

test/list.test.js
```diff
@@ -112,4 +112,10 @@ suite('horizontal', () => {
 		column.textProperty = null;
 		assert.equal(column.getString({ list: ['123', '345', '678']}), '123, 345, 678');
 	});
+
+	test('returns the value of an item', () => {
+		column.valueProperty = 'id';
+		assert.strictEqual(column.getComparableValue({ id: 13 }, 'id'), 13);
+		assert.isUndefined(column.getComparableValue({}, 'nonexisting-path'));
```
cristinecula, line 119: You should also add an assertion for the expected behaviour for this part of the code:

https://github.com/Neovici/cosmoz-omnitable/blob/07ef4946ff1f98437d56088cb4851f077d3a361e/cosmoz-omnitable-column-list.js#L159-L163

dotpointer: @cristinecula

I added a test:

`assert.strictEqual(column.getComparableValue({ items: [{ id: 13 }, { id: 14 }, { id: 15 }] }, 'items'), '13 14 15');`

But it fails, why?

dotpointer: I did a test - I renamed the `getComparableValue` function in `cosmoz-omnitable-column-list.js` and in `list.test.js` the result is that it cannot find the function in the test - so it ignores the list version of the function completely.

megheaiulian:
No longer relevant.

PR #355, opened 2020-04-10
Compute item update effects

nomego:
Only do the necessary recalculations when items in `data` are replaced with new versions


cosmoz-omnitable.js
```diff
@@ -660,12 +667,62 @@ class Omnitable extends translatable(
...
+		if (!Array.isArray(this.columns) || notify == null || notify.path === 'data.length') {
+			return;
+		}
+		if (notify.path !== 'data.splices') {
+			this._setColumnValues();
+			this._debounceFilterItems();
 			return;
 		}
+
 		this._setColumnValues();
```
cristinecula, line 717:
```suggestion
		this._setColumnValues();

		if (notify.path !== 'data.splices') {
			this._debounceFilterItems();
			return;
		}
```

cosmoz-omnitable.js
```diff
@@ -660,12 +667,62 @@ class Omnitable extends translatable(
...
 	}
 
-	_dataChanged() {
-		if (!Array.isArray(this.columns)) {
+	_getItemUpdateEffects(splices) {
+		return splices.reduce((acc, splice) => {
+			if (acc.refilter) {
+				return acc;
+			}
+			const itemsReplaced = splice.type === 'splice' && splice.addedCount === splice.removed.length;
```
cristinecula, line 675: What happens if I splice a number of items, not just replace one? Won't it skip the update entirely?

PR #358, opened 2020-04-29
wip

cristinecula:
Testing instructions:

1. open http://localhost:8000/demo/lit-omnitable.html
you will see a 20000 item omnitable
2. scroll the window
3. sort the entire dataset by clicking on the column headers
4. switch sort direction by clicking same header twice


cristinecula:
This is a demo of the performance aspects of the sorting + rendering of large datasets. Most features are still missing.

PR #374, opened 2020-06-19
fix: use set instead of splice to update items

nomego:


nomego:
Misplaced the error, this should not be changed!

PR #384, opened 2020-09-18
fix: autocomplete, only render spinner when loading #24260

dotpointer:
Check that the `cosmoz-omnitable-column-autocomplete` only shows the loading spinner animation when loading, otherwise the area should not be occupied.

RM:24260.


dotpointer:
@JockeCK is this something you could review? It is quite small.

JockeCK (approved):

PR #385, opened 2020-09-21
fix: cosmoz-omnitable, adding empty set message slot #23679

dotpointer:
Adding empty set message slot.

RM:23679.

### DIFF
--- a/cosmoz-omnitable-styles.js
+++ b/cosmoz-omnitable-styles.js
@@ -121,6 +121,16 @@ container.innerHTML = `<dom-module id="cosmoz-omnitable-styles">
 				padding-bottom: 0;
 			}
 
+			.tableContent-empty div.tableContent-empty-message {
+				@apply --layout-vertical;
+				@apply --layout-center-justified;
+				padding-bottom: 24px;
+			}
+
+			.tableContent-empty.overlay div.tableContent-empty-message {
+				padding-bottom: 0;
+			}
+
 			.tableContent-empty p {
 				font-size: 15px;
 				color: #ddd;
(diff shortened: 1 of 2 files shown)


nomego (commented):
This change seems a bit weird, to have a slot with content where you can inject content.. don't they both get rendered?

dotpointer:
> This change seems a bit weird, to have a slot with content where you can inject content.. don't they both get rendered?

The intent was to use the data in the slot as the default content. How shall it be done to achieve this?

cristinecula:
@dotpointer what happens when you use the slot in the parent component? Is the default content replaced?

dotpointer:
> @dotpointer what happens when you use the slot in the parent component? Is the default content replaced?

Yes this is how this PR works.

The slot data is passed from `upload.html.js`, through the parent slot in `list-core.html.js` to the slot in `cosmoz-omnitable.js`.

If you mean to only use the parent slot located in `upload.html.js` without this pull request with the slot in cosmoz-omnitable then this does not result in the upload text being shown.

nomego:
Ok let's merge this if it visually addresses the issue, and let it be something to support/consider for the lit-omnitable version.
@dotpointer will you make sure the tests pass?

dotpointer:
> Ok let's merge this if it visually addresses the issue, and let it be something to support/consider for the lit-omnitable version.
> @dotpointer will you make sure the tests pass?

Yes.

This seems to be the test result:

FAILED TESTS:
  visible
    ✖ adjusts columns when visible
      Firefox 80.0 (Ubuntu 0.0.0)
    expected true to be false

      + expected - actual

      -true
      +false

I checked the test at `basic.test.js`, why is it failing, I have not changed the visible columns?

dotpointer:
I ran the tests locally, the test did not fail for me.

dotpointer:
@nomego I restarted the tests, they do now seem to pass.

PR #386, opened 2020-09-22
fix: list, only render spinner when loading #24260

dotpointer:
Check that the `cosmoz-omnitable-column-list` only shows the loading spinner animation when loading, otherwise the area should not be occupied.

For example the Assigned to... column in Open invoices in frontend.

RM:24260.


dotpointer:
@JockeCK here is another similar like the previous one.

JockeCK (approved):

PR #390, opened 2020-10-03
Feature/389 remove iron flex layout

### LINKED ISSUE #389: Replace `iron-flex-layout` with flex css
Replace `iron-flex-layout` with flex css.

megheaiulian:
- Update deps
- Replace iron-flex-layout with simple css
- Restore more of the already broken demos
- Adds a resize/visible test (related to #388 )

fixes #389

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,13 +57,13 @@
 		}
 	},
 	"dependencies": {
-		"@neovici/cosmoz-autocomplete": "^2.4.6",
+		"@neovici/cosmoz-autocomplete": "^2.4.9",
 		"@neovici/cosmoz-bottom-bar": "^3.1.4",
 		"@neovici/cosmoz-datetime-input": "^3.0.3",
 		"@neovici/cosmoz-grouped-list": "^3.1.0",
 		"@neovici/cosmoz-i18next": "^3.1.1",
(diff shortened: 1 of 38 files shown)


package.json
```diff
@@ -79,32 +78,31 @@
 		"@polymer/paper-spinner": "^3.0.0",
 		"@polymer/polymer": "^3.3.1",
 		"@webcomponents/shadycss": "^1.10.0",
-		"file-saver": "github:plumelo/FileSaver.js",
-		"web-animations-js": "^2.3.2"
```
nomego, line 83: I think we had to have this as a dep because of the group/sort on menus and the animation it made when opening.. but maybe that's not needed anymore?
Can you just make sure of that?

nomego (approved):
Can we also fix the single eslint warning? :)

package.json
megheaiulian: Removing `web-animations-js` triggers warnings in tests and demos. Because it's in `devDependencies` we can keep it for a while ... ?

megheaiulian: `paper-dropdown-menu` uses this animations.

nomego: Well I mean it should stay as a dep then

PR #392, opened 2020-10-12
fix(removeItem): arrayDelete is undefined

cristinecula:
Broken in https://github.com/Neovici/cosmoz-omnitable/pull/388: by dropping the use of `mixinBehaviors`, it inadvertently removed the arrayDelete method. `mixinBehaviors` applies Polymer's `LegacyElementMixin` under the hood, which includes the arrayDelete method.

Re https://github.com/Neovici/cosmoz-frontend/pull/2804

### DIFF
--- a/test/basic.test.js
+++ b/test/basic.test.js
@@ -274,6 +274,23 @@ suite('item update effects', () => {
 		assert.isTrue(sortSpy.called, 'resorted');
 	});
 
+	test('removeItem removes a given item from the table', () => {
+		const
+			numItems = omnitable.data.length,
+			item = omnitable.data[2];
+
+		assert.equal(omnitable.data.indexOf(item), 2);
+
+		omnitable.removeItem(item);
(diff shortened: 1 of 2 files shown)


test/basic.test.js
```diff
@@ -274,6 +274,18 @@ suite('item update effects', () => {
...
+			numItems = omnitable.data.length,
+			item = omnitable.data[2];
+
+		assert.equal(omnitable.data.indexOf(item), 2);
+
+		omnitable.removeItem(item);
+		assert.equal(omnitable.data.length, numItems - 1);
+		assert.equal(omnitable.data.indexOf(item), -1);
+	});
+
```
nomego, line 288:
```js
test('removeItem gracefully fails to remove a non-existing item from the table', () => {
  assert.equal(omnitable.removeItem({}), null);
});
```

nomego (approved):

megheaiulian (approved):

PR #393, opened 2020-10-14
fix: always selectable if data is valid

megheaiulian:
Don't depend on hasActions from cosmoz-bottom-bar anymore since it does not
make sense to have a omnitable without selection.


megheaiulian:
If/when this gets merged changes from Neovici/cosmoz-bottom-bar#87 could be reverted and bottom bar does not need to notify hasActions.

cristinecula (approved):

PR #407, opened 2021-01-05
feat: adds emptyLabel and emptyValue to list columns

### LINKED ISSUE #406: Support filtering for empty values in list columns
`list` columns should have `emptyLabel` property to add a filter for empty value. 
When set it enables filter for empty (`[]`, `''`, `null`, `undefined`) values.

megheaiulian:
Adds `emptyLabel` and `emptyValue` to list columns allowing filtering for empty values.

Fixes #406


cristinecula (approved):
Great!

PR #409, opened 2021-01-27
fix(column-autocomplete): handle numbers in getComparableValue

### LINKED ISSUE #408: Implement numeric-sort column attribute
attribute to determine alphabetical or numerical method of sorting on autocomplete column

*edit* Iulian found that this feature is probably not needed. 
"It looks like the autocomplete column has a getComparableValue  function that forces it to be String."

Removing the override should suffice

megheaiulian:
Updated the `getComparableValue` fn in `autocomplete` column to handle textProperty and valueProperty.
If the cell value is not an array it returns the value as it is.
Because of that numeric sort should just work.

Fixes #408  but doesn't need to add the attribute.


JockeCK (approved):
tested the code on a few views. worked sorting-wise and let you filter via chips , great work!

PR #416, opened 2021-02-19
feat: custom render group function

### LINKED ISSUE #65: Add support for a custom group row template
The cell template is used to render group rows. The item bound to this template is the first item of the group.
If the cell template (column's default or custom) use an item's property that is not common to the grouped items, it will render something specific to the first item.

One should add support for a specific group template to handle such situations.

cristinecula:
Fixes #65


megheaiulian (approved):
Great that it also fixes #65

PR #230, opened 2018-10-01
Feature: Show total number of rows

Wurper:
If totalNumberOfRows is used, total number of rows will be shown in footer.

PR #421, opened 2021-03-23
feat(cosmoz-omnitable): renders totalAvailable in row stats

megheaiulian:
Adds a `totalAvailable` property that is used to render the total available
number of items if greater than current data.
#230

### DIFF
--- a/cosmoz-omnitable.js
+++ b/cosmoz-omnitable.js
@@ -1585,7 +1585,7 @@ class Omnitable extends mixin({ isEmpty }, getEffectiveChildrenLegacyMixin(trans
 
 	_renderRowStats(numRows, totalAvailable) {
 		if (Number.isInteger(totalAvailable) && totalAvailable > numRows) {
-			return this.ngettext('{0} / {1} row', '{0} / {1} rows', numRows, totalAvailable);
+			return this.ngettext('{1} / {0} row', '{1} / {0} rows', totalAvailable, numRows);
 		}
 		return this.ngettext('{0} row', '{0} rows', numRows);
 	}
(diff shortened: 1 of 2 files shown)


cosmoz-omnitable.js
```diff
@@ -1582,6 +1582,13 @@ class Omnitable extends mixin({ isEmpty }, getEffectiveChildrenLegacyMixin(trans
 	_debounce(name, fn, asyncModule = timeOut.after(0)) {
 		this.debouncers[name] = Debouncer.debounce(this.debouncers[name], asyncModule, fn);
 	}
+
+	_renderRowStats(numRows, totalAvailable) {
+		if (Number.isInteger(totalAvailable) && totalAvailable > numRows) {
+			return this.ngettext('{0} / {1} row', '{0} / {1} rows', numRows, totalAvailable);
```
cristinecula, line 1588: How is the singular selected in this case? 
Is it determined by the last parameter?
If not, it might display `1 / 250 row`

test/basic.test.js
```diff
@@ -480,3 +480,13 @@ suite('render group function', () => {
...
 });
 
+suite('render row stats', () => {
+	test('renders totalAvailable', async () => {
+		const omnitable = await setupOmnitableFixture(html`
+			<cosmoz-omnitable .totalAvailable=${ 2000 }>
+				<cosmoz-omnitable-column name="name" value-path="name">
+				</cosmoz-omnitable-column>
+			</cosmoz-omnitable>`, generateTableDemoData(10, 10, 25));
+		assert.equal(omnitable.shadowRoot.querySelector('.footer-tableStats span:last-child').textContent, '{0} / {1} rows');
```
cristinecula, line 490: Shouldn't you assert actual values?
`100 / 2000 rows`

cristinecula (changes requested):
Great! A couple of questions:

cristinecula (commented):
Perfect!

PR #423, opened 2021-03-24
fix: migrate header-row to haunted

cristinecula:
Re https://github.com/Neovici/cosmoz-frontend/issues/3282


cristinecula:
TODO:

* [ ] header alignment no longer works properly

megheaiulian (approved):

Issue #426, opened 2021-03-31
Deterministic header layout algorithm


cristinecula:
More than half of the initial rendering time for a new omnitable is spent in calculating the layout (determining if columns fit/toggling columns). The current algorithm leads to the ugly "column dance", that users experience when the data loads.

I propose writing a simple algorithm that mimics the flex layout behavior and determines the proper column widths, in a single pass. It will also determine which columns should be visible/hidden, based on the priority settings.

cristinecula:
This will also fix https://github.com/Neovici/cosmoz-omnitable/issues/317

PR #427, opened 2021-04-01
feat: fast-layout

cristinecula:
New attribute `fast-layout` that switches the omnitable to the new, more performant, layout algorithm. This will allow us to test the new algorithm in alpha, before fully switching to it. I expect that adjustments will need to be made in most of the list-cores, because the new algorithm does not take into account the cell content, whereas the old algo used the first rendered row as a reference for sizing the columns.

Fixes https://github.com/Neovici/cosmoz-omnitable/issues/426


megheaiulian (approved):
Amazing job!

cristinecula:
Thank you! :heart:

PR #429, opened 2021-04-16
feat: implement multi-select

### LINKED ISSUE #177: multi-select

megheaiulian:
Implements multi-select:
- ctrl -> select only this item/group
- shift + click -> toggle a range from last interacted

fixes #177


cristinecula (approved):
Great work. I tried it out and it feels nice.
I have only one issue: if you shift click just outside the checkbox, it makes this ugly text selection. Maybe we can improve on this in the future.

PR #431, opened 2021-04-20
Drop legacy layout

cristinecula:
BREAKING CHANGE: layout algorithm has been changed to fast-layout. Some differences in rendering will occur.

Also included:
* migrate item-expand to haunted
* migrate item-expand-line to haunted
* migrate @apply css mixin to css parts


cristinecula:
package-lock needs re-do

JockeCK (approved):
tested on the demos.

PR #448, opened 2021-08-25
Finish resizable columns implementation

cristinecula:
Also included:
* deprecated visibleColumns and disabledColumns
* fix editable attribute being set on the cells, even though column is not editable
* fix for expand rows not resizing when columns are collapsed/expanded

### DIFF
--- a/cosmoz-omnitable-column-mixin.js
+++ b/cosmoz-omnitable-column-mixin.js
@@ -118,11 +118,6 @@ export const columnMixin = dedupingMixin(base => class extends base { // eslint-
 				value: '75px'
 			},
 
-			resizeWidth: {
-				type: Number,
-				observer: () => { /* HACK: force polymer to listen for `resizeWidth` property changes */ }
-			},
-
 			/**
 			 * Base width of this column when in edit mode.
 			 */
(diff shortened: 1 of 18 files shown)


cosmoz-omnitable-column-mixin.js
```diff
@@ -456,7 +451,7 @@ export const columnMixin = dedupingMixin(base => class extends base { // eslint-
 	_propertiesChanged(currentProps, changedProps, oldProps) {
 		super._propertiesChanged(currentProps, changedProps, oldProps);
 		// let the repeaters know that this column has been updated and that it needs to re-render the templates
-		this.dispatchEvent(new CustomEvent('cosmoz-column-prop-changed', { bubbles: true, detail: { currentProps, changedProps, oldProps }}));
+		this.dispatchEvent(new CustomEvent('cosmoz-column-prop-changed'));
```
megheaiulian, line 454: We don't need currentProps, changedProps etc... anymore ?

megheaiulian (approved):

cosmoz-omnitable-column-mixin.js
cristinecula: No, it's not used anywhere.

PR #475, opened 2021-12-10
Fix render issue in cosmoz-data-nav

cristinecula:
Caught during integration with Cosmoz frontend. When navigating a cosmoz-data-nav, the omnitables that are rendered off-screen do not properly render the list. The <iron-list> skips rendering if it is not connected.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
     branches:
       - master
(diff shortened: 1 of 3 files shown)


cristinecula:
Postpone merging a bit, this introduces some performance issues.

cristinecula:
Performance issue fixed.

megheaiulian (approved):

PR #486, opened 2022-01-13
fix: use null for filter reset

megheaiulian:
Use null for filter reset so that Polymer trigger change notifications correctly.


cristinecula (approved):
Thank you!

PR #487, opened 2022-01-17
fix(default-column): workaround for double-encoded filters

cristinecula:
Re RM:27661

### DIFF
--- a/cosmoz-omnitable-column-mixin.js
+++ b/cosmoz-omnitable-column-mixin.js
@@ -85,6 +85,13 @@ export const
 		}
 
 		deserializeFilter(column, filter) {
+			if (filter == null) {
+				return null;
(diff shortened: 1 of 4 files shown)


JockeCK (changes requested):
test/basic.test.js:

 ❌ basic > deserializeFilter returns object
      AssertionError: expected '[object Object]' to deeply equal { key: 'value' }
        at o.<anonymous> (test/basic.test.js:76:10)

 ❌ basic > deserializeFilter handles null
      AssertionError: expected 'null' to equal null
        at o.<anonymous> (test/basic.test.js:80:10)

cristinecula:
I've improved the code and fixed the tests.

JockeCK (approved):
great!

PR #502, opened 2022-04-11
fix(sort-and-group-options):  fallback descending/groupByDescending

megheaiulian:
Uses the `descending` and `groupByDescending` defined on the host when
the hash param are empty.


cristinecula (approved):
Great

PR #523, opened 2022-05-31
Upgrade to cosmoz-grouped-list v4 and onItemClick bugfix

cristinecula:
Latest grouped-list has major performance improvements and enables conversion of cosmoz-omnitable to lit.
Also includes fix for: when the omnitable is grouped and you click a row, the emitted index is the index of the virtual row, not of the actual item.


megheaiulian (approved):

cristinecula:
More changes are needed.

PR #540, opened 2022-08-22
fix: refactor to haunted component 

cristinecula:
Fixes AB#7355


cristinecula:
I've finished refactoring the old code and dropped the use of a base class entirely. I will test it by linking it with FE and it should be ready for merging.

cristinecula:
@megheaiulian I've tested the integration of FE with this version of omnitable and I caught a minor bug and fixed it. It is ready for review.

megheaiulian (approved):

cristinecula:
I will revert the v12 changes and prepare a v13 release with this instead.

PR #542, opened 2022-10-06
fix(use-settings): error when removing columns

cristinecula:
An error occurs when columns with custom settings (size, order, etc) are dropped from the columns list.

Fixes https://neovici.slack.com/archives/C6LJQMJFM/p1664996455187809?thread_ts=1664913886.064969&cid=C6LJQMJFM


JockeCK (approved):
So the ol [filter(Boolean) trick](https://michaeluloth.com/filter-boolean). 

(I googled it, did not know about it)

And now I see I was too slow.

PR #552, opened 2022-11-08
fix(styles): shim styles

megheaiulian:
Fix CSS not having Shaddy css shim applied.

Note: We should replace most Polymer components and drop the shim requirements.
Without it  content is not properly styled in some cases.


cristinecula (commented):
Nice catch!

PR #553, opened 2022-11-12
fix(use-public-interface): memoize

megheaiulian:
Memoize public interference.

### DIFF
--- a/lib/use-public-interface.js
+++ b/lib/use-public-interface.js
@@ -67,7 +67,13 @@ const mkNapi = (host) => {
 	};
 };
 
-export const usePublicInterface = ({ host, visibleData, columns, ...api }) => {
+/*eslint no-unused-vars: ["error", { "ignoreRestSiblings": true }]*/
+export const usePublicInterface = ({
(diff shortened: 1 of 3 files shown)


lib/use-public-interface.js
```diff
@@ -35,32 +22,61 @@ export const usePublicInterface = ({ host, visibleData, ...api }) => {
...
+export const usePublicInterface = ({
+	host,
+	visibleData,
+	...api
+}) => {
+	const { setFilterState } = api,
+		napi = useMemo(() => mkNapi(host), []);
+
+	useImperativeApi(api, Object.values(api));
+	useImperativeApi(napi, Object.values(napi));
```
cristinecula, line 79:
```suggestion
	useImperativeApi(napi, []);
```
Because napi values never change

megheaiulian: No. This makes it clear that napi would be re-applied when its values change. The values will never change because of the useMemo above but that is a separate thing.

cristinecula (approved):

PR #554, opened 2022-11-15
fix(drodpown): correct placement

megheaiulian:
Correctly set the placement property instead of the attribute.


cristinecula (approved):
:smile:

Issue #92, opened 2017-11-22
Datetime column filter can be different values


nomego:
When set by the user, it will be a date, but when reloading with URL filter persist, it will be set as an ISO String.

It should always be a date.

megheaiulian:
This is closed by #98

PR #559, opened 2023-01-31
feat(treenode column): passing along three new values

JockeCK:
Needed for treenode column change [#92](https://github.com/Neovici/cosmoz-omnitable-treenode-column/pull/92)

### DIFF
--- a/lib/use-dom-columns.js
+++ b/lib/use-dom-columns.js
@@ -62,6 +62,7 @@ const columnSymbol = Symbol('column'),
 			width: column.width,
 			flex: column.flex,
 			priority: column.priority,
+			limit: column.limit,
 
 			getString: column.getString,
 			getComparableValue: column.getComparableValue,
@@ -116,9 +117,8 @@ const columnSymbol = Symbol('column'),
 			// treenode columns
 			ownerTree: column.ownerTree,
 			keyProperty: column.keyProperty,
-			limit: column.limit,
-			headerTextProperty: column.headerTextProperty,
-			headerValueProperty: column.headerValueProperty,
+			hideFromRoot: column.hideFromRoot,
+			showMaxNodes: column.showMaxNodes,
 
 			[columnSymbol]: column,
 		}));
--- a/lib/use-dom-columns.js
+++ b/lib/use-dom-columns.js
@@ -62,7 +62,6 @@ const columnSymbol = Symbol('column'),
 			width: column.width,
 			flex: column.flex,
 			priority: column.priority,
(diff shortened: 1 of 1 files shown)


megheaiulian (commented):
@cristinecula It sucks that we need to do this. 
Couldn't we have a way for dom columns to expose their own properties that should be passed to render? Maybe the renderCell, renderHeader etc.. should have easy access the full column object.

lib/use-dom-columns.js
```diff
@@ -116,6 +117,8 @@ const columnSymbol = Symbol('column'),
 			// treenode columns
 			ownerTree: column.ownerTree,
 			keyProperty: column.keyProperty,
+			hideFromRoot: column.hideFromRoot,
+			showMaxNodes: column.showMaxNodes,
```
megheaiulian, line 122: To avoid listing all the possible properties of all columns in one place we should refactor this a bit to call a optional function implementation `getConfig` on each column. 

```suggestion
...column.getConfig?.(column),
```

In that case in treenode columns we can implement this: 
```
 getConfig(column) {
 return {
limit: column.limit,
hideFromRoot: column.hideFromRoot
....etc
}
}

```

megheaiulian (changes requested):

lib/use-dom-columns.js
JockeCK: ok, pushed change.

lib/use-dom-columns.js
```diff
@@ -62,6 +62,7 @@ const columnSymbol = Symbol('column'),
 			width: column.width,
 			flex: column.flex,
 			priority: column.priority,
+			limit: column.limit,
```
megheaiulian, line 65: We can remove this if we have column.getConfig

lib/use-dom-columns.js
```diff
@@ -116,6 +117,10 @@ const columnSymbol = Symbol('column'),
 			// treenode columns
 			ownerTree: column.ownerTree,
 			keyProperty: column.keyProperty,
+			hideFromRoot: column.hideFromRoot,
```
megheaiulian, line 120: and this too

lib/use-dom-columns.js
megheaiulian: @JockeCK Can we remove this lines ? only `...column.getConfig?.(column),` is needed.

JockeCK: Of course, sorry for late replies.

megheaiulian (approved):

PR #560, opened 2023-02-03
feature/AB9162

CarloBar1:
Task [AB#9162](https://dev.azure.com/neovici/Cosmoz3/_workitems/edit/9162/).
Change the aspect of omnitable headers so to look like this:
image

### DIFF
--- a/cosmoz-omnitable-column-amount.js
+++ b/cosmoz-omnitable-column-amount.js
@@ -119,6 +119,7 @@ class OmnitableColumnAmount extends columnMixin(PolymerElement) {
 		source
 	) {
 		return html`<cosmoz-omnitable-amount-range-input
+			class="editable-header"
 			.title=${ title }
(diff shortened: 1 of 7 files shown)


CarloBar1:
The font-family attribute added doesn't seem to affect the placeholders in the header, just the "Boolean" field shown in the local development.

CarloBar1:
Another update: I tried putting various font specifications around for the code, and in no case it was rendered, always keeping the default font.

cristinecula:
Please rebase and squash the commits. Make sure that CI tests pass.

cristinecula:
Can you find a way to make the font styles easier to edit from outside of the omnitable?
Either exposing the headers themselves as parts or using some css variables?
I think that it would make sense to be able to set `--cosmoz-omnitable-header-font` to customize all of the headers.

PR #561, opened 2023-02-03
link-underline-removal

CarloBar1:
Task [AB#9329](https://dev.azure.com/neovici/Cosmoz3/_workitems/edit/9329/).
The underline on links should be shown only when the mouse hovers them.

The expected final behaviour is shown in the video:
https://user-images.githubusercontent.com/122988480/217528578-108df213-2dad-4ea3-9dba-e5895ddf03e7.mp4

### DIFF
--- a/cosmoz-omnitable-styles.js
+++ b/cosmoz-omnitable-styles.js
@@ -83,11 +83,14 @@ export default css`
 		position: relative;
 		overflow: hidden;
 	}
-
 	:host a {
 		color: var(--primary-link-color, inherit);
+		text-decoration: var(--cosmoz-omnitable-link-decoration, none);
 	}
-
+	:host a:hover {
+		color: var(--primary-link-color, inherit);
+		text-decoration: var(--cosmoz-omnitable-link-decoration, underline);
+	}
 	/* The wrapping div that contains the header, the table content and the footer */
 	.mainContainer {
 		background-color: #fff;
@@ -125,6 +128,7 @@ export default css`
 	.header > cosmoz-omnitable-header-row {
 		flex: auto;
 	}
+
 	.header-cell {
 		--paper-input-container: {
 			padding-top: 0;
@@ -136,6 +140,7 @@ export default css`
 	cosmoz-omnitable-header-row {
 		white-space: nowrap;
 	}
+
 	cosmoz-omnitable-header-row > div {
 		display: inline-block;
 		box-sizing: border-box;
@@ -155,10 +160,8 @@ export default css`
 		width: 7px;
 		height: 30px;
 		margin-left: -3px;
-
 		background: transparent;
 		cursor: ew-resize;
-
 		z-index: 1000;
 		user-select: none;
 	}
@@ -189,6 +192,9 @@ export default css`
 
 	cosmoz-omnitable-item-row > div[hidden] {
 		display: none !important;
+	}
+	cosmoz-omnitable-item-row > div a {
+		text-decoration: none;
 	}
 
 	.tableContent {
@@ -258,7 +264,6 @@ export default css`
 	.tableContent-empty.overlay div.tableContent-empty-message {
 		padding-bottom: 0;
 	}
-
 	.tableContent-empty p {
 		font-size: 15px;
 		color: #ddd;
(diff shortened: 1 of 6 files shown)


cristinecula:
Please include reference to issue.

cosmoz-omnitable-styles.js
```diff
@@ -86,9 +86,11 @@ export default css`
 	//The following code makes the link underlining appear only on mouseover
 	:host a {
 		color: var(--primary-link-color, inherit);
+		text-decoration: none;
 	}
 	:host a:hover {
 		color: var(--primary-link-color, inherit);
+		color: lightblue;
```
cristinecula, line 93: Are you sure you want to override the color style here?

cristinecula (changes requested):

cosmoz-omnitable-styles.js
```diff
@@ -258,7 +267,6 @@ export default css`
 	.tableContent-empty.overlay div.tableContent-empty-message {
 		padding-bottom: 0;
 	}
-
```
cristinecula, line 261: SO many useless whitespace changes. Are they added by prettier? If not, please revert them.

cosmoz-omnitable-styles.js
```diff
@@ -190,6 +195,10 @@ export default css`
 	cosmoz-omnitable-item-row > div[hidden] {
 		display: none !important;
 	}
+	cosmoz-omnitable-item-row > div a {
+		text-decoration: none;
+		color: grey;
```
cristinecula, line 200: If you are going to set a color, you should use a variable, to enable customization.

cristinecula (changes requested):

cosmoz-omnitable-styles.js
cristinecula: Is setting a color even necessary here?

cosmoz-omnitable-styles.js
CarloBar1: I applied Prettier and it separates every selector's brackets with a whitespace - I just re-applied Prettier over the whole document now

cosmoz-omnitable-styles.js
```diff
@@ -83,11 +83,15 @@ export default css`
 		position: relative;
 		overflow: hidden;
 	}
-
+	//The following code makes the link underlining appear only on mouseover
```
cristinecula, line 86: Comment is just explaining the obvious.
```suggestion
```

cristinecula (changes requested):
I would like to have this configurable. Can we add `--cosmoz-omnitable-link-decoration` and `--cosmoz-omnitable-link-decoration--hover` variables?

cosmoz-omnitable-styles.js
CarloBar1: Comment deleted, decoration variabilized

cosmoz-omnitable-styles.js
```diff
@@ -190,6 +193,9 @@ export default css`
 	cosmoz-omnitable-item-row > div[hidden] {
 		display: none !important;
 	}
+	cosmoz-omnitable-item-row > div a {
+		text-decoration: none;
```
cristinecula, line 197: This should also use the css var.

cosmoz-omnitable-styles.js
```diff
@@ -83,11 +83,14 @@ export default css`
 		position: relative;
 		overflow: hidden;
 	}
-
 	:host a {
 		color: var(--primary-link-color, inherit);
+		text-decoration: var(--cosmoz-omnitable-link-decoration, none);
```
cristinecula, line 88: Please keep the underline, as that is the default browser behaviour. We'll style it differently in frontend.
```suggestion
		text-decoration: var(--cosmoz-omnitable-link-decoration, underline);
```

cristinecula (changes requested):

cosmoz-omnitable-styles.js
```diff
@@ -190,6 +193,9 @@ export default css`
 	cosmoz-omnitable-item-row > div[hidden] {
 		display: none !important;
 	}
+	cosmoz-omnitable-item-row > div a {
+		--cosmoz-omnitable-link-decoration: none;
+	}
```
cristinecula, line 198:
```suggestion
```

This sets the `text-decoration` to none, which is the opposite of what we want.

cristinecula (changes requested):

cristinecula (approved):

PR #562, opened 2023-02-08
fix(settings): keep collapsed columns hidden when disabling a column

cristinecula:
Fixes [AB#8399](https://dev.azure.com/neovici/7e94365d-32b1-416f-854b-7f195da31da6/_workitems/edit/8399)

### DIFF
--- a/lib/settings/cosmoz-omnitable-settings.js
+++ b/lib/settings/cosmoz-omnitable-settings.js
@@ -49,7 +49,6 @@ const placement = ['bottom-right', ...defaultPlacement],
 					.checked=${checked}
 					@click=${onToggle}
 					.indeterminate=${indeterminate}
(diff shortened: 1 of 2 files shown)


lib/settings/cosmoz-omnitable-settings.js
```diff
@@ -49,7 +49,6 @@ const placement = ['bottom-right', ...defaultPlacement],
 					.checked=${checked}
 					@click=${onToggle}
 					.indeterminate=${indeterminate}
-					.windeterminate=${indeterminate}
```
megheaiulian, line 52: This is not a typo!

cristinecula: I know, but it's not needed anymore.

PR #564, opened 2023-02-08
Feature 9162 rebased

CarloBar1:
Rebased branch for [previous pull request](https://github.com/Neovici/cosmoz-omnitable/pull/560), as there were too many wrong commit messages

### DIFF
--- a/cosmoz-omnitable-styles.js
+++ b/cosmoz-omnitable-styles.js
@@ -132,7 +132,33 @@ export default css`
 			--paper-font-caption_-_line-height: 18px;
 		}
 	}
+	cosmoz-omnitable-amount-range-input,
+	cosmoz-omnitable-number-range-input,
+	cosmoz-omnitable-date-range-input,
+	cosmoz-omnitable-time-range-input,
+	cosmoz-omnitable-datetime-range-input,
+	paper-input {
+		--paper-font-subhead_-_font-family: var(
+			--cosmoz-omnitable-header-font-family,
+			'Inter' sans-serif
+		);
+		text-transform: var(--cosmoz-omnitable-header-text-transform, uppercase);
+		--paper-font-subhead_-_font-weight: var(
+			--cosmoz-omnitable-header-font-weight,
+			bold
+		);
+		--paper-font-subhead_-_font-size: var(
+			--cosmoz-omnitable-header-font-size,
+			14pt
+		);
+	}
 
+	cosmoz-autocomplete-ui::part(input-label) {
+		text-transform: var(--cosmoz-omnitable-header-text-transform, uppercase);
+		font-weight: var(--cosmoz-omnitable-header-font-weight, bold);
+		font-family: var(--cosmoz-omnitable-header-font-family, 'Inter' sans-serif);
+		font-size: var(--cosmoz-omnitable-header-font-size, 14pt);
+	}
 	cosmoz-omnitable-header-row {
 		white-space: nowrap;
 	}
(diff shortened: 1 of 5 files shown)


cosmoz-omnitable-styles.js
```diff
@@ -132,7 +132,28 @@ export default css`
...
+	cosmoz-omnitable-number-range-input,
+	cosmoz-omnitable-date-range-input,
+	cosmoz-omnitable-time-range-input,
+	cosmoz-omnitable-datetime-range-input,
+	paper-input {
+		--paper-font-subhead_-_font-family: var(
+			--cosmoz-header-font-family,
+			'Inter',
+			sans-serif
+		);
```
cristinecula, line 145: Are you sure that this works correctly? AFAIK `var` only takes 2 parameters, and you have 3.

cosmoz-omnitable-styles.js
```diff
@@ -132,7 +132,28 @@ export default css`
...
+		text-transform: var(--cosmoz-header-text-transform, uppercase);
```
cristinecula, line 146: You should include the component name in the css var, so it is clear when someone is reading the CSS the purpose of a certain variable.

```suggestion
		text-transform: var(--cosmoz-omnitable-header-text-transform, uppercase);
```

cristinecula (changes requested):
Almost there.

cosmoz-omnitable-styles.js
CarloBar1: In the [CSS var() definition](https://developer.mozilla.org/en-US/docs/Web/CSS/var), it says that a var declaration can allow more than one fallback value after the first comma. It is though a bit unclear whether the fallback values are considered one after each other, so I just deleted those commas.

cristinecula: Please figure out how it is supposed to be written and update accordingly. Experiment with both variants and figure out which is doing the right thing.

cosmoz-omnitable-styles.js
```diff
@@ -132,7 +132,33 @@ export default css`
 			--paper-font-caption_-_line-height: 18px;
 		}
 	}
+	cosmoz-omnitable-amount-range-input,
+	cosmoz-omnitable-number-range-input,
+	cosmoz-omnitable-date-range-input,
+	cosmoz-omnitable-time-range-input,
+	cosmoz-omnitable-datetime-range-input,
+	paper-input {
```
cristinecula, line 140: It should work the same if you defined these styles for the header containing element.
```suggestion
	[selector-for-header-container] {
```

Please try it out.

lib/settings/cosmoz-omnitable-settings.js
```diff
@@ -144,8 +144,9 @@ const placement = ['bottom-right', ...defaultPlacement],
 				html`<div class="headerDots" slot="button">...</div>` :
 				html`
 					<svg viewBox="0 0 24 24" width="24" slot="button" fill="currentColor">
-						<path d="M10 18h5V5h-5v13zm-6 0h5V5H4v13zM16 5v13h5V5h-5z"></path>
-					</svg>`
+					<circle cx="10" cy="18" r="2"></circle>
+						<circle cx="10" cy="12" r="2"></circle>
+						<circle cx="10" cy="6" r="2"></circle></svg>`
```
cristinecula, line 149: Please post a screenshot.

cristinecula (changes requested):

lib/settings/cosmoz-omnitable-settings.js
CarloBar1: image

CarloBar1: Right now it looks like this:
image

cosmoz-omnitable-styles.js
CarloBar1: Included all in the .header-cell selector, and it works.

PR #566, opened 2023-02-21
CSS Vars reset in Omnitable Header styling

CarloBar1:
Re [AB#9162](https://dev.azure.com/neovici/7e94365d-32b1-416f-854b-7f195da31da6/_workitems/edit/9162) - Follow-up of a previous [pull request](https://github.com/Neovici/cosmoz-omnitable/pull/564)- the styling settings in the omnitable folder are reset to previous defaults, but they are made modifiable with the usage of CSS vars.

The screenshot shows the desired outcome:
image

### DIFF
--- a/cosmoz-omnitable-styles.js
+++ b/cosmoz-omnitable-styles.js
@@ -132,25 +132,24 @@ export default css`
 			--paper-font-caption_-_line-height: 18px;
 		}
 		--paper-font-subhead_-_font-family: var(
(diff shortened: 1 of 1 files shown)


cristinecula (changes requested):
image

I checked out the v12.7.2 version and in that one the rendered font is Roboto. 

image
image

Please determine what is the correct default value so it matches the previous version.

cristinecula (approved):

PR #567, opened 2023-02-21
Applying trim for more lenient searching

JockeCK:
This is for when users c&p something for example and an extra spacebar at the beginning an/or end sneaks in. 

Added an test ('_applySingleFilter returns true if filterString is equal to value despite extra space characters added') as well. (the rest is formatting.)


megheaiulian (approved):
The change looks fine to me and its really nice that you added tests.

:+1:

PR #590, opened 2023-08-22
fix(boolean column): element name mismatch fix

JockeCK:
small fix, just noticed it in the code


megheaiulian (approved):
I  see now what you mean. It's mismatch between opening and closing tags. :facepalm:

PR #592, opened 2023-08-30
feat: render data error in omnitable

hassanalbaaj:
Render error in omnitable based on sent in value

[AB#11960](https://dev.azure.com/neovici/Cosmoz3/_workitems/edit/11960/)

image (1)

### DIFF
--- a/test/render-list.test.js
+++ b/test/render-list.test.js
@@ -1,0 +1,14 @@
+import { renderList } from '../lib/render-list';
+import { expect, html, fixture, nextFrame } from '@open-wc/testing';
+
+suite('render-list', () => {
+	test('error', async () => {
+		const el = await fixture(
+			html` <div class="tableContent" id="tableContent">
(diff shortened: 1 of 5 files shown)


megheaiulian (changes requested):
Please add a snapshot test for this. 
Just call render-list with a error and put that into a snapshot.

test/render-list.test.js
```diff
@@ -0,0 +1,14 @@
+import { renderList } from '../lib/render-list';
+import { expect, html, fixture, nextFrame } from '@open-wc/testing';
+
+suite('render-list', () => {
+	test('error', async () => {
+		const el = await fixture(
+			html` <div class="tableContent" id="tableContent">
```
megheaiulian, line 7: Why do we need this wrapper? 
Can't we just use
```
 const el = await fixture(renderList({error: new Error('whatever')}))
```
?

megheaiulian (approved):

PR #593, opened 2023-09-11
fix: added a minimum width to the status columns

CarloBar1:
Feature AB#10366 - set a minimum width for the status columns in omnitable. 
image

### DIFF
--- a/cosmoz-omnitable-styles.js
+++ b/cosmoz-omnitable-styles.js
@@ -53,9 +53,7 @@ const checkbox = css`
 	}
 
 	.checkbox:hover {
-		box-shadow:
-			0 0 0 2px rgba(0, 0, 0, 1) inset,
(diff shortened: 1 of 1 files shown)


cristinecula:
Two problems:

1. This will not work properly with the column resize algorithm. All widths need to be defined in JS, we can't use CSS for this. There are three properties that can be configured on the columns: width, minWidth and flex. These are the only way to control the width of the columns. 
2. cosmoz-omnitable should not be making such decisions. Defining a min width for columns named status or invoiceStatus has no place in this repo, which is a public component, meant to be usable in any project.

The correct solution is to change this in the tables configuration, in the views of cosmoz-frontend. Due to code duplication and legacy issues, this unfortunately means editing each and every place where a status column is defined. Please create a new PR in cosmoz-frontend, in which you set width (or minWidth) on the status columns.

PR #594, opened 2023-09-20
fix(lib/use-dom-columns): fallback sortOn, groupOn on valuePath

megheaiulian:
Uses valuePath that fallbacks to name for sortOn & groupOn.


cristinecula (commented):
Nice addition!

PR #609, opened 2024-04-19
fix(dependencies): new inputs needed in all their instances

CarloBar1:
About Feature [AB#11519](https://dev.azure.com/neovici/7e94365d-32b1-416f-854b-7f195da31da6/_workitems/edit/11519) - latest cosmoz-input release (v5.0.0) needs to be installed in all the components using it.


cristinecula (changes requested):
Looking at the lock diff, it also requires the latest versions of @neovici/cosmoz-autocomplete and @neovici/cosmoz-datetime-input

cristinecula (approved):

PR #629, opened 2024-09-24
fix: reset to factory omnitable

basilgood:
https://dev.azure.com/neovici/Cosmoz3/_workitems/edit/16820

### DIFF
--- a/lib/settings/use-saved-settings.js
+++ b/lib/settings/use-saved-settings.js
@@ -37,7 +37,7 @@ export default (settingsId, settings, setSettings, onReset) => {
 				}
 				onReset?.();
 			},
-			[onReset]
+			[onReset],
 		),
 		hasChanges: settings != null,
 	};
--- a/lib/settings/drivers/remote.js
+++ b/lib/settings/drivers/remote.js
@@ -8,7 +8,7 @@ export default ({ prefix = '', get$, post$ } = {}) => {
 		}
 	};
 	return {
-		write: async (settingsId, settings) => {
+		write: async (settingsId, settings = {}) => {
 			const key = prefix + settingsId;
 			try {
 				await post$(key, settings);


lib/settings/use-saved-settings.js
```diff
@@ -31,13 +31,13 @@ export default (settingsId, settings, setSettings, onReset) => {
...
+				setSavedSettings();
 				if (e.shiftKey) {
-					await write(settingsId);
-					setSavedSettings();
+					await write(settingsId, {});
 				}
 				onReset?.();
 			},
-			[onReset]
+			[settings, savedSettings, write, settingsId],
```
cristinecula, line 40: `settings` and `savedSettings` are not dependencies of this callback, but `onReset` is.

```suggestion
			[onReset, write, settingsId],
```

lib/settings/use-saved-settings.js
```diff
@@ -31,13 +31,13 @@ export default (settingsId, settings, setSettings, onReset) => {
 		onReset: useCallback(
 			async (e) => {
 				setSettings();
+				setSavedSettings();
```
cristinecula, line 34: 1. Setting both `settings` and `savedSettings` to `undefined` will make it so the omnitable uses the default layout, but if the shift key is not pressed, this state will be temporary, because the `savedSettings` are not written to storage. Upon refresh, the user will see again the previously `savedSettings`. 

2. We don't want to always reset the `savedSettings`, but only when the shift key is pressed, so we "reset to factory". The user might have a saved layout and does some changes. In that case "reset" should reset to the saved layout, not to the default layout.

lib/settings/use-saved-settings.js
```diff
@@ -31,13 +31,13 @@ export default (settingsId, settings, setSettings, onReset) => {
...
 				if (e.shiftKey) {
-					await write(settingsId);
-					setSavedSettings();
+					await write(settingsId, {});
```
cristinecula, line 36: This is the correct fix, but it should be done in the "driver".

```suggestion
					await write(settingsId);
```

The `local` driver handles `write` being called with a single argument, by using `localStorage.removeItem` to clean up the stored state.
https://github.com/Neovici/cosmoz-omnitable/blob/dfd7a85e1eaa66a577efd509cf3465433ee41bcf/lib/settings/drivers/local.js

You should do something similar in the `remote` driver, as that is the one with the bug (the API does not support `undefined`, but expects an empty object).
https://github.com/Neovici/cosmoz-omnitable/blob/dfd7a85e1eaa66a577efd509cf3465433ee41bcf/lib/settings/drivers/remote.js

cristinecula (changes requested):
Good work investigating this! You have found the source of the issue, but the fix should be applied in the `driver`.

lib/settings/use-saved-settings.js
basilgood: Issue says Reset to factory settings and not reset settings and factory settings. I have no way of knowing what you want to treat if you don't specify exactly.

cristinecula: All I mean is that there is no need for this change.

cristinecula (approved):

cristinecula (changes requested):
Please rename commit to use `fix:` not `chore:`

cristinecula (approved):

PR #630, opened 2024-10-22
fix(lib): title align to the right + use font-variant-numeric

amandaavalencia:
Omnitable title align to the right and use font-variant-numeric for date
[AB#17293](https://dev.azure.com/neovici/Cosmoz3/_workitems/edit/17293/)

### DIFF
--- a/lib/cosmoz-omnitable-number-range-input.js
+++ b/lib/cosmoz-omnitable-number-range-input.js
@@ -1,7 +1,7 @@
-import { _ } from '@neovici/cosmoz-i18next';
 import { PolymerElement } from '@polymer/polymer';
-import { html } from 'lit-html';
 import { ifDefined } from 'lit-html/directives/if-defined.js';
+import { html } from 'lit-html';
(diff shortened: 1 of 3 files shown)


lib/cosmoz-omnitable-number-range-input.js
```diff
@@ -58,11 +64,7 @@ class NumberRangeInput extends rangeInputMixin(
 				?opened=${this.headerFocused}
 				@opened-changed=${onOpenedChanged}
 			>
-				<div
-					class="dropdown-content"
-					slot="dropdown-content"
-					style="padding: 15px; min-width: 100px;"
-				>
+				<div class="dropdown-content classen" slot="dropdown-content">
```
micaelagoffe, line 67: Is there a reason why you don't use the class already defined? `dropdown-content` that is 🙂

amandaavalencia: no reason! just forgot to change this, thanks for the comment :)

micaelagoffe (approved):
LGTM!

PR #634, opened 2024-12-09
fix(footer): set max-toolbar-items when mini table is used

micaelagoffe:
Set max-toolbar-items on cosmoz-bottom-bar when mini table is used

[AB#17737](https://dev.azure.com/neovici/Cosmoz3/_backlogs/backlog/UX/Stories/?workitem=17737)

### DIFF
--- a/lib/render-footer.js
+++ b/lib/render-footer.js
@@ -11,61 +11,56 @@ export const renderFooter = ({
 	csvFilename,
 	xlsxFilename,
 	xlsxSheetname,
-	topPlacement
-}) => {
-	return html`
-		<cosmoz-bottom-bar
-			id="bottomBar"
-			?active=${!isEmpty(selectedItems.length)}
-		>
-			<slot name="info" slot="info">
-				${ngettext(
-					'{0} selected item',
-					'{0} selected items',
-					selectedItems.length
-				)}
-			</slot>
-			<slot name="actions" id="actions"></slot>
-			<!-- These slots are needed by cosmoz-bottom-bar
+	topPlacement,
+	maxToolbarItems,
+}) =>
+	html`<cosmoz-bottom-bar
+		id="bottomBar"
+		?active=${!isEmpty(selectedItems.length)}
+		?max-toolbar-items=${!isEmpty(maxToolbarItems)}
+	>
+		<slot name="info" slot="info">
+			${ngettext(
+				'{0} selected item',
+				'{0} selected items',
+				selectedItems.length,
+			)}
+		</slot>
+		<slot name="actions" id="actions"></slot>
+		<!-- These slots are needed by cosmoz-bottom-bar
         as it might change the slot of the actions to distribute them in the menu -->
-			<slot name="bottom-bar-toolbar" slot="bottom-bar-toolbar"></slot>
-			<slot name="bottom-bar-menu" slot="bottom-bar-menu"></slot>
-			<cosmoz-dropdown-menu slot="extra" .placement=${topPlacement}>
-				<svg
-					slot="button"
-					width="14"
-					height="18"
-					viewBox="0 0 14 18"
-					fill="none"
-					stroke="currentColor"
-					xmlns="http://www.w3.org/2000/svg"
-				>
-					<path
-						d="M1 8.5L7.00024 14.5L13 8.5"
-						stroke-width="2"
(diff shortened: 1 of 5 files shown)


lib/use-footer.js
```diff
@@ -1,22 +1,21 @@
...
-		} = host;
+		} = host,
+		isMiniTable = host.hasAttribute('mini');
 
 	return {
 		csvFilename,
 		xlsxFilename,
 		xlsxSheetname,
 		topPlacement,
+		isMiniTable,
```
cristinecula, line 18:
```suggestion
		maxToolbarItems,
```

lib/use-footer.js
```diff
@@ -1,22 +1,21 @@
...
-}) => {
+export const useFooter = ({ host, ...rest }) => {
 	const {
 			csvFilename = 'omnitable.csv',
 			xlsxFilename = 'omnitable.xlsx',
 			xlsxSheetname = 'Omnitable',
 			topPlacement = _defaultPlacement,
-		} = host;
+		} = host,
+		isMiniTable = host.hasAttribute('mini');
```
cristinecula, line 11: This is not a great way to determine if the omnitable is in `mini` state. You should make use of [isMiniSize](https://github.com/Neovici/cosmoz-omnitable/blob/18c0da7d4dc8ffa8e11708f0fd529ab0271715b5/lib/use-mini.js#L5) and determine the number of toolbar items in the main hook.

```suggestion
      maxToolbarItems,
		} = host;
```

cristinecula (changes requested):

cristinecula:
Ideally the cosmoz-bottom-bar element would use intersection observers or some other mechanism to determine if it should show/hide the button.

lib/use-omnitable.js
```diff
@@ -40,14 +40,15 @@ export const useOmnitable = (host) => {
...
-		{ collapsedColumns, miniColumns } = useFastLayout({
+		{ isMiniSize, collapsedColumns, miniColumns } = useFastLayout({
 			host,
 			columns,
 			settings,
 			setSettings,
 			resizeSpeedFactor,
 			sortAndGroupOptions,
 		}),
+		maxToolbarItems = host.hasAttribute('mini') && isMiniSize ? '0' : '',
```
cristinecula, line 51: Reading the `isMini` state from the `host` element is too indirect and will lead to confusion. Instead, you should alter the code in  `useMini` to provide this information as a flag.

https://github.com/Neovici/cosmoz-omnitable/blob/18c0da7d4dc8ffa8e11708f0fd529ab0271715b5/lib/use-mini.js#L3-L29

```js
  const hasMiniColumn =  !!miniColumn

	useEffect(() => {
		host.toggleAttribute('mini', hasMiniColumn);
	}, [hasMiniColumn]);

  return {
    isMini: hasMiniColumn && isMiniSize
  }
```

cristinecula:
```suggestion
		maxToolbarItems = isMini ? 0 : undefined, // should be a number or undefined
```

lib/render-footer.js
```diff
@@ -11,61 +11,56 @@ export const renderFooter = ({
...
-			</slot>
-			<slot name="actions" id="actions"></slot>
-			<!-- These slots are needed by cosmoz-bottom-bar
+	topPlacement,
+	maxToolbarItems,
+}) =>
+	html`<cosmoz-bottom-bar
+		id="bottomBar"
+		?active=${!isEmpty(selectedItems.length)}
+		?max-toolbar-items=${!isEmpty(maxToolbarItems)}
```
cristinecula, line 20: `?prop` should only be used for boolean attributes, for number props you should use `.prop`. 

```suggestion
		.maxToolbarItems=${maxToolbarItems}
```

cristinecula (changes requested):

lib/use-omnitable.js
micaelagoffe: After our discussion in slack I'm using ´'0' : undefined´, so I can keep the isEmpty check in render-footer.

lib/render-footer.js
micaelagoffe: Keeping this change after our discussion in slack 🙂

cristinecula (approved):

PR #635, opened 2024-12-10
fix: emit event for when mini table is used

micaelagoffe:
Emit event with useNotifyProperty for when mini table is used.

[AB#17737](https://dev.azure.com/neovici/Cosmoz3/_workitems/edit/17737/)


cristinecula (approved):
LGTM!

PR #662, opened 2025-07-09
feat: border radius on header row

fredrikbernholm:


alexseman (approved):
Would be nice to have a tab indentation at lines `557`, `560` & `565`.

PR #674, opened 2025-08-31
feat: upgrade dependencies

fredrikbernholm:
Looking for things that we're not applied in omnitable. Not sure we should or need to do this though. Lots of changes...


megheaiulian:
Closed as I don't think its applicable anymore.

Issue #670, opened 2025-08-15
Upgrade paper-dropdown-menu to cosmoz-dropdown


megheaiulian:
Omnitable header column use paper-dropdown-menu to show range inputs in a dropdown.
We should replace the dropdown itself with cosmoz-dropdown based implementations.
We can start with number first.

megheaiulian:
We will probably need a customized version cosmoz-dropdown to use for the headers.

megheaiulian:
Implemented in related PRs. We need to tweak the dropdown a bit or probably implement another input based on cosmoz-input to act as the trigger.

PR #675, opened 2025-09-01
Feature/cosmoz dropdown

jhay-25:
Replaces paper-dropdown-menu with cosmoz-dropdown https://github.com/Neovici/cosmoz-omnitable/issues/670

### DIFF
--- a/src/lib/cosmoz-omnitable-range-input-mixin.js
+++ b/src/lib/cosmoz-omnitable-range-input-mixin.js
@@ -350,7 +350,7 @@ export const rangeInputMixin = (base) =>
 				100,
 			);
 		}
-		
+
 		/**
 		 * Debounced function called by `_filterInputChanged` when `_filterInput` changes.
 		 *


jhay-25:
Reusing the hooks from `cosmoz-dropdown` was making cosmoz-input buggy (e.g it wouldn’t let you highlight input values).

I recreated the hooks in this commit:
[https://github.com/Neovici/cosmoz-omnitable/pull/675/commits/87b51f5043da436cf8d1fd3c4dd9847eb07208f2](https://github.com/Neovici/cosmoz-omnitable/pull/675/commits/87b51f5043da436cf8d1fd3c4dd9847eb07208f2?utm_source=chatgpt.com)
Everything works as expected there

lib/components/cosmoz-omnitable-dropdown-menu/cosmoz-omnitable-dropdown-menu.js
```diff
@@ -0,0 +1,234 @@
...
+};
+
+const Header = (host) => {
+	const {
+		label,
+		placeholder,
+		title,
+		horizontalAlign,
+		placement,
+		class: className,
```
cristinecula, line 157:
```suggestion
```

You don't need to relay the class this way. You can use `:host(selector)` instead.
https://developer.mozilla.org/en-US/docs/Web/CSS/:host_function

cristinecula (changes requested):

megheaiulian (changes requested):
Need to be rebased,  re-reviewed and we need to figure out a  implementation that does not duplicate too much of cosmoz-dropdown.

PR #686, opened 2025-09-16
fix: omnitable content visual bug

fredrikbernholm:
Fixes visual bug reported in https://github.com/Neovici/cosmoz-frontend/pull/8942


cristinecula:
The "working set empty" message is hidden underneath the tableContent-scroller element, which now has a white background.

Screenshot_2025-09-17_08-25-06
Screenshot_2025-09-17_08-24-52

PR #692, opened 2025-09-24
Converted utils-amount.js to typescript

jhay-25:


megheaiulian:
@jhay-25 Please keep as draft until ready for review :pray:

megheaiulian (changes requested):
Need to be rebased ...

jhay-25:
Applied the reviewed changes here
https://github.com/Neovici/cosmoz-omnitable/pull/696#discussion_r2397524374

src/lib/utils-amount.ts
```diff
@@ -0,0 +1,200 @@
...
+
+const convertToAmount = (value: Amount): Amount | null => {
+	const number: unknown = toNumber(value.amount);
+
+	if (typeof number === 'number') {
+		return { currency: value.currency, amount: number };
+	}
+
+	return null;
+};
```
megheaiulian, line 33: This is the proper way to  implement type checking with custom predicates. 
Unfortunately at first we need to the conversion as is and not modify so much of the sources as it makes it harder to review.
Lets keep this as is in `toAmount` and refactor later on.

megheaiulian (changes requested):

src/lib/utils-amount.ts
jhay-25: updated

megheaiulian (approved):

PR #696, opened 2025-10-01
Converted utils-number.js to typescript

jhay-25:
Converted utils-number.js to typescript.

### DIFF
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -2,50 +2,24 @@ export type LimitFunction = (a: number, b: number) => number | undefined;
 
 export type Currency = 'SEK' | 'USD' | 'EUR' | 'AUD';
 
-export interface AmountValue {
+export interface Amount {
 	amount: string | number;
 	currency: Currency | string;
 }
 
-export type ValuePath = 'amount' | 'min' | 'max' | 'value';
+export type GetPath = string | (string | number)[];
 
-export interface Value {
-	valuePath?: ValuePath;
+export interface NumberColumn {
+	valuePath?: GetPath;
 	locale?: string;
 	minimumFractionDigits?: number | null;
 	maximumFractionDigits?: number | null;
 }
 
-export interface AmountObj {
-	amount: AmountValue;
-	bool: boolean;
-	date: Date;
-	dateJSON: string;
-	group: string;
-	id: number;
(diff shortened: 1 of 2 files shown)


src/lib/types.ts
```diff
@@ -0,0 +1,25 @@
+export type LimitFunction = (a: number, b: number) => number | undefined;
+
+export type Currency = 'SEK' | 'USD' | 'EUR' | 'AUD';
+
+export interface Amount {
+	amount: string | number;
+	currency: Currency | string;
+}
```
cristinecula, line 8: I don't think you need the `Currency` type.

`Currency | string` ends up as `string` 
See [playground](https://www.typescriptlang.org/play/#code/FAUwHgDg9gTgLgAjgTwiBBhArjGIB2AxsggLwIDkAygKIDSFCAPpQKpUAijLFNrASt0oBBVlwDcwUJFiIAlvjggYAMwCGhdMIC2ULIoQBvYAEg1u-XABcCAM5wYCgObME+LNoBGyyScI48ImQbbFwCYld7R3wnSQBfKRQ0BAANMgQdPUUAbQp-MKCKAF1gAHpShEqAPQB+IA)

Image

The default typescript types just use `string` for currency (Intl.NumberFormatter):

```ts
    interface NumberFormatOptions {
        localeMatcher?: "lookup" | "best fit" | undefined;
        style?: NumberFormatOptionsStyle | undefined;
        currency?: string | undefined;
        currencyDisplay?: NumberFormatOptionsCurrencyDisplay | undefined;
        useGrouping?: NumberFormatOptionsUseGrouping | undefined;
        minimumIntegerDigits?: number | undefined;
        minimumFractionDigits?: number | undefined;
        maximumFractionDigits?: number | undefined;
        minimumSignificantDigits?: number | undefined;
        maximumSignificantDigits?: number | undefined;
    }
```

Also, I'm not so sure about `Amoun['amount']` being `string | number`. Shouldn't we enforce just `number`?

```ts
export interface Amount {
	amount: number;
	currency: string;
}
```

jhay-25: https://github.com/Neovici/cosmoz-omnitable/pull/692/commits/a5920cc107288d88f7a059dbf18a2a2db220cd98

PR #723, opened 2025-10-15
Converted use-collapsible-items.js to typescript

jhay-25:


src/grouped-list/use-collapsible-items.ts
```diff
@@ -1,10 +1,11 @@
 import { useCallback } from '@pionjs/pion';
 import { useWeakState } from './use-weak-state';
 import { isGroup } from './utils';
+import { Item } from '../lib/types';
```
megheaiulian:
```suggestion
import type { Item } from '../lib/types';
```

megheaiulian (approved):

Issue #765, opened 2025-10-31
Implement a more standard dropdown trigger for range columns


megheaiulian:
## What:
Implement a more standard dropdown trigger for range columns.
In #670 we replaced `paper-dropdown-menu` with `cosmoz-dropdown`.
We styled the dropdown to look similar and for the trigger/anchor we used some custom markup and CSS to make it look the same as the other  inputs used in the headers.
Instead we should use input based on `cosmoz-input`'s render function  + CSS to make it look and work the same.

## Why
We had to put quite a bit of style to make it look like a cosmoz-input and we basically re-invented the label floating behavior.
Need to reuse as much as possible. 

## How

First we need to export `render` and `style` from the `cosmoz-input` component's package.
Then in omnitable  we need to create a `cosmoz-omnitable-dropdown-input` element using the `render` and `style` of `cosmoz-input` but instead of using a actual input as control we can use a simple div. 
The in all dropdown we can reuse that element and only style the dropdown removing some default styles.
We should be reusing all the floating and focusing behavior and style.

Depends on Neovici/cosmoz-input#126

PR #775, opened 2025-11-04
Use cosmoz-input as dropdown trigger

jhay-25:
Use cosmoz-input as dropdown trigger https://github.com/Neovici/cosmoz-omnitable/issues/765

### DIFF
--- a/src/lib/cosmoz-omnitable-dropdown-input.js
+++ b/src/lib/cosmoz-omnitable-dropdown-input.js
@@ -19,12 +19,19 @@ const style = css`
 		--text-align: center;
 	}
 
+	:host::part(control) {
+		width: 100%;
+	}
+
 	#input {
 		color: #737373;
 		user-select: none;
 		cursor: pointer;
 		min-height: var(--line-height);
 		text-align: var(--text-align);
+		overflow: hidden;
+		text-overflow: ellipsis;
+		white-space: nowrap;
 	}
 
 	label {


src/lib/cosmoz-omnitable-dropdown-input.js
```diff
@@ -0,0 +1,87 @@
...
+import { html } from 'lit-html';
+import { render, styles } from '@neovici/cosmoz-input';
+
+const observedAttributes = ['label', 'value', 'slot', 'text-align', 'focused'];
+
+const DropdownInput = (host) => {
+	const { label, value, slot, textAlign, focused } = host;
+
+	if (textAlign) {
+		host.style.setProperty('--text-align', textAlign);
```
megheaiulian: We usually avoid setting styles this way.

src/lib/cosmoz-omnitable-dropdown-input.js
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	useEffect(() => {
```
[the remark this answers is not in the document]
megheaiulian: Do not worry about that. We need to make all this work in cosmoz-input with a external variable. 
When we will set a --input-focused variable on the input it apply the same style as if focused .

src/lib/cosmoz-omnitable-dropdown-input.js
```diff
@@ -0,0 +1,65 @@
...
+
+	return render(control, { label });
+};
+
+customElements.define(
+	'cosmoz-omnitable-dropdown-input',
+	component(DropdownInput, {
+		observedAttributes,
+		styleSheets: [
+			sheet(styles),
```
megheaiulian: We can use the css tagged template literal from @pionjs.

Before DropdownInput we can start define
``` js
import {styles as inputStyles} from '@neovici/cosmoz-input';
const style = css`
${inputStyles}
:host {}
...
`
```

src/lib/cosmoz-omnitable-dropdown.js
```diff
@@ -49,145 +67,28 @@ export const renderDropdown = ({
...
 		>
-			<div slot="button" class="dropdown-button">${title || tooltip}</div>
-
-			${filterText
-				? html`<div slot="button" class="filter-value">${filterText}</div>`
-				: nothing}
-
-			<div class="dropdown-content" @mousedown=${(e) => e.stopPropagation()}>
+			<cosmoz-omnitable-dropdown-input
+				input-focused=${headerFocused}
```
cristinecula: You should use a boolean attribute.

```suggestion
				?input-focused=${ifDefined(headerFocused)}
```

src/lib/cosmoz-omnitable-dropdown-input.js
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	label {
+		display: block;
+		text-align: var(--text-align);
+	}
+
+	:host([input-focused='true']) {
+		--focused: focused;
+	}
```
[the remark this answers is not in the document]
megheaiulian: I don't think we even need that at all. From outside we can always just set `style="--focused: focused"`

megheaiulian (approved):

Issue #772, opened 2025-11-03
Rename sort-on on to sort-path


megheaiulian:
# Rename sort-on on to sort-path

We currently use sort-on (with fallback to value-path) as the configuration that tell us what property to use when sorting items locally.
We should rename it to `sort-path` to make it clearer that this is the property used for local sorting and so that its name does not conflict with `sort-on` on omnitable itself.

PR #777, opened 2025-11-06
Renamed sort-on to sort-path

jhay-25:
Renamed sort-on propert on individual columns to sort-path https://github.com/Neovici/cosmoz-omnitable/issues/772

### DIFF
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -20,8 +20,8 @@ export interface Column {
 	priority?: number;
 	flex?: string;
 	width?: string;
-	sortOn?: boolean;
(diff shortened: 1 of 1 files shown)


megheaiulian:
We have a sortOn property on cosmoz-omnitable itself. 
This property contains the name of the column we currently sort on. 
We also have a sortOn property on each column. This one contains the path in data that it should be used for sorting. 
That is confusing, as self evident by our talks in the PR. 
That was why I thought we should name the property that is set on the column to sort-path

cristinecula (approved):

Issue #773, opened 2025-11-03
Implement `no-sort` on columns


megheaiulian:
#Implement `no-sort` on columns 

Implement `no-sort` attribute/property on omnitable columns. 
When set to `true` a column is not sortable:
 - it does not display any  sort ui in the header
 - it does not show up in the Sort section in omnitable settings
 - if sort from hash param is set to that column it should be ignored.

PR #784, opened 2025-11-11
Implement no-sort on column

jhay-25:
Implemented no-sort on column https://github.com/Neovici/cosmoz-omnitable/issues/773

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -129,9 +129,9 @@
       }
     },
     "node_modules/@babel/helper-validator-identifier": {
-      "version": "7.27.1",
-      "resolved": "https://registry.npmjs.org/@babel/helper-validator-identifier/-/helper-validator-identifier-7.27.1.tgz",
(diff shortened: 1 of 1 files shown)


src/lib/use-sort-and-group-options.js
```diff
@@ -82,6 +82,12 @@ export const useSortAndGroupOptions = (
 		// eslint-disable-next-line no-void
 		useEffect(() => void (resetRef.current = setSG), []);
 
+		useEffect(() => {
+			if (sortOn && sortOnColumn?.noSort) {
+				setSortOn(undefined);
+			}
+		}, [sortOn, sortOnColumn, setSortOn]);
```
megheaiulian: I don't think this is needed. We can consider noSort as more of a ui setting. 
So if  a noSort column has no UI to sort that's enough for us even if programmatically you could still set it.

demo/hash-param.html
```diff
@@ -19,7 +19,8 @@
 		<script type="module">
 			import { html } from 'lit-html';
 			import { component } from '@pionjs/pion';
-			import { hashbang } from '@neovici/cosmoz-page-router';
+			import { hashbang } from '@neovici/cosmoz-router';
```
megheaiulian: I think only one import is enough ?

Issue #774, opened 2025-11-03
Move sort handle in the input's labels


megheaiulian:
#Move sort handle in the input's labels
Move sort handle in the input's labels so that:
 1. UI does not jump around on hover
 2. The sort icon floats together with the label

Somewhat dependant on #765

PR #791, opened 2025-11-19
Moved sort handle beside dropdown label

jhay-25:
https://github.com/Neovici/cosmoz-omnitable/issues/774


megheaiulian (changes requested):
We need to replace the current symbol/graphics we use for sort currently with a thin/narrow arrow.
We need to place that inside the label of all header column inputs. The sort symbol should float with the label, have a hover state and point up or down (or be hidden) depending on the current sort.

Issue #804, opened 2025-11-28
NEO-217 Select ALL items


cristinecula:
# Context

Users may need to perform bulk operations on very large datasets. The omnitable should enable users to perform actions on *all* items in a dataset, not just those loaded locally.

To support this, we need a “Select ALL” feature that represents operating on all items.

# Solution

When a user selects all rows, show in the footer: *“50 selected items. \[Select all matching items\]”.*

Clicking `Select all matching items`  changes the text to: *"All matching items selected."*

### **Inspiration Screenshots (Gmail)**

Empty selection

<img width="1554" height="195" alt="Image" src="https://github.com/user-attachments/assets/35922146-66ab-4c7f-8996-b8dd2d3fb4b4" />

Select all local items

<img width="1556" height="268" alt="Image" src="https://github.com/user-attachments/assets/d6181b40-8b90-4d0d-900b-acf0cb78f024" />

Select all items

<img width="1557" height="270" alt="Image" src="https://github.com/user-attachments/assets/4d3b247f-58ac-4f8b-9a01-721bee60a385" />

# Implementation

- [ ] define and export a symbol `All`
- [ ] add \[Select all matching items\] button to the footer
- [ ] handle clicks by making `selectedItems = All`
- [ ] when `selectedItems === All` all rows should be rendered as selected
- [ ] add `enable-select-all` attribute that determines the display of the \[Select all matching items\] button

https://linear.app/neovici/issue/NEO-217/select-all-items

PR #817, opened 2025-12-04
Added select-all feature

jhay-25:
https://github.com/Neovici/cosmoz-omnitable/issues/804


cristinecula:
Tried it out and feels very nice!

A couple of issues I noticed:

1. when ALL items are selected, the 100 selected items text should be removed

expected
Screenshot_2025-12-11_10-50-54

actual
Screenshot_2025-12-11_10-50-34

2. the select all checkbox should render as selected when either the entire dataset is selected or when ALL items are selected

Screenshot_2025-12-11_10-51-10

cristinecula (changes requested):
It works well, but I think there is a styling issue: the checkboxes do not render properly.

Image

After checking, I see that the issue is also present in main: https://neovici.github.io/cosmoz-omnitable/?path=/docs/components-comsmozomnitablefulldemo--documentation

I saw that the issue is fixed in your other PR: https://github.com/Neovici/cosmoz-omnitable/pull/825

Image

So I guess that we should get that one merged first and then you can rebase this one.

PR #818, opened 2025-12-05
fix: column config is invalidated every time the contents of the slot changes

cristinecula:
In [NEO-311](https://linear.app/neovici/issue/NEO-311/list-queue-infinite-load-loop) it has become evident that the column config is being recomputed whenever ANYTHING is added/removed to the slot. This causes unnecessary re-renders because everything is recomputed when the config changes, including the `visibleData`.

As mentioned there:

> We should only recalculate the config if the element that was
added/removed from the slot is an omnitable column. The problem is that the `slotchange` event does not tell us this information. We need to keep track of the `assignedNodes` ourselves.


cristinecula:
I will investigate the test errors later today or tomorrow.

Issue #824, opened 2025-12-09
Improve the default style of omnitable


megheaiulian:
We need to improve the default style of omnitable and match it more  to the style in Cosmoz itself.
We should also take the oportunity to cleanup the old Polymer CSS.

PR #825, opened 2025-12-10
Improved the default style of omnitable 

jhay-25:
https://github.com/Neovici/cosmoz-omnitable/issues/824


src/lib/settings/style.css.js
```diff
@@ -194,7 +194,7 @@ export default css`
 	.sg[data-on] {
 		background: var(
 			--cosmoz-omnitable-checkbox-checked-color,
-			var(--primary-color)
+			var(--primary-color, #508aef)
```
megheaiulian: We will need to introduce a `--comoz-omnitable-primary-color` variable and use that instead of --primary-color everywhere.

megheaiulian (changes requested):

PR #1037, opened 2026-07-05
feat: add select all items feature

basilgood:
## Description

Adds a "Select all items" feature (Gmail-style) for paginated datasets, where only a subset of items may be loaded while a total count is known.

### Changes

- **`All` sentinel** — a `Symbol('All')` that represents "all items" (the full dataset, not just the loaded subset)
- **`enable-select-all` attribute** — boolean attribute on `<cosmoz-omnitable>` to enable the feature
- **`allItemsCount` property** — allows consumers to pass the total count (e.g., from API's `totalAvailable`) when only a subset is loaded
- **Footer UI** — shows "Select all items" button when some items are selected; shows "All {count} items selected" (or "All items selected" when no count is provided) after clicking
- **Header checkbox** — unchanged, selects only loaded items as array
- **Selection logic** — `isItemSelected` / select / deselect all handle the `All` sentinel correctly

### Storybook

Added `SelectAll` and `BulkSelectAll` stories with 100 items loaded and `allItemsCount: 10000` to demonstrate the paginated scenario.

Closes FE-66

### DIFF
--- a/src/grouped-list/use-selected-items.ts
+++ b/src/grouped-list/use-selected-items.ts
@@ -1,5 +1,6 @@
 import { useCallback, useEffect, useProperty, useState } from '@pionjs/pion';
 import type { Item } from '../lib/types';
+import { All, type TAll } from '../lib/utils';
 import { GroupItem, isGroup } from './utils';
 
 export interface UseSelectedItemsParams {
@@ -9,7 +10,7 @@ export interface UseSelectedItemsParams {
 }
 
 export interface UseSelectedItemsResult {
-	selectedItems: Item[];
+	selectedItems: Item[] | TAll;
 	isItemSelected: (item: Item) => boolean;
 	isGroupSelected: (group: GroupItem<Item>) => boolean;
 	isSelected: (item: Item | GroupItem<Item>) => boolean;
@@ -27,7 +28,7 @@ export const useSelectedItems = ({
 	data,
 	flatData,
 }: UseSelectedItemsParams): UseSelectedItemsResult => {
-	const [selectedItems, setSelectedItems] = useProperty<Item[]>(
+	const [selectedItems, setSelectedItems] = useProperty<Item[] | TAll>(
 		'selectedItems',
 		() => [],
 	);
@@ -36,15 +37,16 @@ export const useSelectedItems = ({
 	 * Check if item is selected.
 	 */
 	const isItemSelected = useCallback(
-		(item: Item) => selectedItems.includes(item),
+		(item: Item) => selectedItems === All || selectedItems.includes(item),
 		[selectedItems],
 	);
 	/**
 	 * Check if group is selected.
 	 */
 	const isGroupSelected = useCallback(
-		(group: GroupItem<Item>) => group?.items?.every(isItemSelected) ?? false,
-		[isItemSelected],
+		(group: GroupItem<Item>) =>
+			selectedItems === All || (group?.items?.every(isItemSelected) ?? false),
+		[selectedItems, isItemSelected],
 	);
 	/**
 	 * Check if item.group is selected.
@@ -60,23 +62,36 @@ export const useSelectedItems = ({
 	const select = useCallback((item: Item | GroupItem<Item>) => {
 		const groupItem = item as GroupItem<Item>;
 		const items = groupItem.items ?? [item as Item];
-		setSelectedItems((selection) => [
-			...selection,
-			...items.filter((i: Item) => !selection.includes(i)),
-		]);
+		setSelectedItems((selection) => {
+			if (selection === All) {
+				return selection;
+			}
+			return [
+				...selection,
+				...items.filter((i: Item) => !selection.includes(i)),
+			];
+		});
 		setLastSelection(item);
 	}, []);
 	/**
 	 * Removes an item/group from the list of selected items.
 	 */
-	const deselect = useCallback((item: Item | GroupItem<Item>) => {
-		const groupItem = item as GroupItem<Item>;
-		const items = groupItem.items ?? [item as Item];
-		setSelectedItems((selection) =>
-			selection.filter((i) => !items.includes(i)),
-		);
-		setLastSelection(item);
-	}, []);
+	const deselect = useCallback(
+		(item: Item | GroupItem<Item>) => {
+			const groupItem = item as GroupItem<Item>;
+			const items = groupItem.items ?? [item as Item];
+			setSelectedItems((selection) => {
+				if (selection === All) {
+					return (flatData ?? []).filter(
+						(selectedItem) => !items.includes(selectedItem as Item),
+					) as Item[];
+				}
+				return selection.filter((i) => !items.includes(i));
+			});
+			setLastSelection(item);
+		},
+		[flatData],
+	);
 
 	const selectOnly = useCallback((item: Item | GroupItem<Item>) => {
 		const groupItem = item as GroupItem<Item>;
@@ -136,7 +151,7 @@ export const useSelectedItems = ({
 	useEffect(
 		() =>
 			setSelectedItems((selectedItems) =>
-				selectedItems.length > 0 && flatData
+				selectedItems !== All && selectedItems.length > 0 && flatData
 					? (flatData.filter((i) =>
 							selectedItems.find((item) => compareItemsFn(i, item)),
 						) as Item[])
--- a/src/grouped-list/use-selected-items.ts
+++ b/src/grouped-list/use-selected-items.ts
@@ -82,9 +82,12 @@ export const useSelectedItems = ({
 			const items = groupItem.items ?? [item as Item];
 			setSelectedItems((selection) => {
 				if (selection === All) {
-					return (flatData ?? []).filter(
-						(selectedItem) => !items.includes(selectedItem as Item),
-					) as Item[];
+					return (flatData ?? [])
+						.filter((selectedItem) => !isGroup(selectedItem))
+						.filter(
+							(selectedItem) =>
+								!items.some((item) => compareItemsFn(selectedItem, item)),
+						) as Item[];
 				}
 				return selection.filter((i) => !items.includes(i));
 			});
--- a/src/grouped-list/use-selected-items.ts
+++ b/src/grouped-list/use-selected-items.ts
@@ -1,6 +1,6 @@
 import { useCallback, useEffect, useProperty, useState } from '@pionjs/pion';
 import type { Item } from '../lib/types';
-import { All, type TAll } from '../lib/utils';
+import { All, type TSelection } from '../lib/utils';
 import { GroupItem, isGroup } from './utils';
 
 export interface UseSelectedItemsParams {
@@ -10,7 +10,7 @@ export interface UseSelectedItemsParams {
 }
 
 export interface UseSelectedItemsResult {
-	selectedItems: Item[] | TAll;
+	selectedItems: TSelection<Item>;
 	isItemSelected: (item: Item) => boolean;
 	isGroupSelected: (group: GroupItem<Item>) => boolean;
 	isSelected: (item: Item | GroupItem<Item>) => boolean;
@@ -28,7 +28,7 @@ export const useSelectedItems = ({
 	data,
 	flatData,
 }: UseSelectedItemsParams): UseSelectedItemsResult => {
-	const [selectedItems, setSelectedItems] = useProperty<Item[] | TAll>(
+	const [selectedItems, setSelectedItems] = useProperty<TSelection<Item>>(
 		'selectedItems',
 		() => [],
 	);
@@ -85,8 +85,7 @@ export const useSelectedItems = ({
 					return (flatData ?? [])
 						.filter((selectedItem) => !isGroup(selectedItem))
 						.filter(
-							(selectedItem) =>
-								!items.some((item) => compareItemsFn(selectedItem, item)),
+							(selectedItem) => !items.includes(selectedItem as Item),
 						) as Item[];
 				}
 				return selection.filter((i) => !items.includes(i));
--- a/src/lib/render-footer.ts
+++ b/src/lib/render-footer.ts
@@ -17,7 +17,7 @@ interface RenderFooterParams {
 	xlsxSheetname?: string;
 	topPlacement?: string;
 	enableSelectAll?: boolean;
-	allSelected?: boolean;
+	allSelected: boolean;
 	allItemsCount?: number;
 }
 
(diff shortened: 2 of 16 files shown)


src/lib/use-footer.ts
```diff
@@ -1,12 +1,25 @@
...
 	host: FooterHost;
+	selectedItems: Item[] | typeof All;
+	setSelectedItems: (
+		items:
+			| Item[]
+			| typeof All
+			| ((prev: Item[] | typeof All) => Item[] | typeof All),
+	) => void;
+	enableSelectAll: boolean;
+	matchingItemsCount: number;
```
megheaiulian: We do not have a matchingItems concept. matching what ? Maybe something like  `allItemsCount`  or something related to selection and or all

src/grouped-list/use-selected-items.ts
```diff
@@ -72,9 +79,12 @@ export const useSelectedItems = ({
 	const deselect = useCallback((item: Item | GroupItem<Item>) => {
 		const groupItem = item as GroupItem<Item>;
 		const items = groupItem.items ?? [item as Item];
-		setSelectedItems((selection) =>
-			selection.filter((i) => !items.includes(i)),
-		);
+		setSelectedItems((selection) => {
+			if (selection === All) {
```
megheaiulian: if selection is `All` and you deselect a group  selection should become all items except everything from that group ?

src/lib/utils.ts
```diff
@@ -2,6 +2,13 @@ const indexSymbol = Symbol('index');
 
 export { indexSymbol };
 
+/**
+ * Sentinel symbol representing "all items matching the current filter".
+ * Set as `selectedItems` when the user clicks "Select all matching items".
+ * The total count is provided via the `totalMatchingItems` property.
+ */
+export const All = Symbol('All');
```
megheaiulian, line 10: We should also export TAll instead of `typeof`ing when we need it

src/lib/render-footer.ts
```diff
@@ -4,33 +4,63 @@ import { html } from 'lit-html';
...
 		part="bottomBar"
 		exportparts="bar: bottomBar-bar, info: bottomBar-info, buttons: bottomBar-buttons"
 	>
-		<slot name="info" slot="info">
-			${t('{count} selected item', { count: selectedItems.length })}
-		</slot>
+		<span slot="info">
+			${selectedItems === All
+				? t('All {count} matching items selected', {
+						count: matchingItemsCount,
```
megheaiulian: We need to handle the case when this count is not provided from outside at all.

src/lib/use-omnitable.js
```diff
@@ -91,7 +91,10 @@ export const useOmnitable = (host) => {
 		footer: useFooter({
 			host,
 			selectedItems,
+			setSelectedItems,
 			columns,
+			enableSelectAll: host.hasAttribute('enable-select-all'),
```
megheaiulian: hasAttribute ?

src/lib/use-omnitable.js
```diff
@@ -91,7 +91,10 @@ export const useOmnitable = (host) => {
...
+			matchingItemsCount: host.totalMatchingItems ?? visibleData.length,
```
megheaiulian: no.

megheaiulian (changes requested):

src/lib/use-omnitable.js
megheaiulian: Why fallback to visibleData.length ?

src/grouped-list/use-selected-items.ts
basilgood: now deselecting one item exits bulk mode and falls back to normal array selection, so the bar shows the regular count again.
eg: all = 100, All = 1000 => All is enabled => deselect one item => bulk button become hidden and bottombar show Selected: 99

src/lib/use-footer.ts
basilgood: renamed matchingItemsCount to allItemsCount

src/lib/utils.ts
basilgood: exported as a named type so we don’t keep repeating typeof All.

src/lib/use-omnitable.js
basilgood: replaced hasAttribute with pion property access (host.enableSelectAll === true).

src/lib/render-footer.ts
basilgood: made the count optional and branching in the footer:
if allItemsCount exists → All {count} items selected
if not → All items selected

src/lib/use-omnitable.js
basilgood: removed the visibleData.length fallback entirely; missing count is handled explicitly in the footer

src/lib/render-footer.ts
```diff
@@ -55,15 +90,19 @@ export const renderFooter = ({
...
 			>
 				${t('Save selected items as CSV')}
 			</button>
 			<button
 				@click=${() =>
-					saveAsXlsxAction(
+					saveAsXlsx(
 						columns as XlsxColumn[],
-						selectedItems,
+						selectedItems === All ? visibleData : selectedItems,
```
megheaiulian: This does not seem correct: when `selectedItems`  is `All`  visibileData is not actually all items.
Maybe we need to hide these when we select all items ? Or handle it differently

megheaiulian (changes requested):

src/lib/render-footer.ts
basilgood: Resolved by hiding the CSV/XLSX export menu when selectedItems === All. When bulk "Select all items" is active, visibleData is not the full dataset, so exporting it would be misleading.

src/grouped-list/use-selected-items.ts
```diff
@@ -27,24 +34,29 @@ export const useSelectedItems = ({
...
-	const [selectedItems, setSelectedItems] = useProperty<Item[]>(
+	const [selectedItems, setSelectedItems] = useProperty<Item[] | TAll>(
 		'selectedItems',
 		() => [],
 	);
 	const [lastSelection, setLastSelection] = useState<Item | GroupItem<Item>>();
+	const dataRef = useRef(data);
+	useEffect(() => {
+		dataRef.current = data;
+	}, [data]);
```
megheaiulian: Why is this need ?

src/grouped-list/use-selected-items.ts
megheaiulian: interesting. Wouldn't it be better to look at flat data ?

.storybook/manager.ts
```diff
@@ -0,0 +1,55 @@
+import { addons } from 'storybook/manager-api';
```
megheaiulian: Why is this needed ?

src/grouped-list/use-selected-items.ts
megheaiulian: Ah you wanna use  it on deselect. I think flatData works better.

src/grouped-list/use-selected-items.ts
megheaiulian: if we look at flat data you don't have to loop thru group items like that.

src/lib/render-footer.ts
```diff
@@ -4,72 +4,112 @@ import { html } from 'lit-html';
...
-						selectedItems,
-						xlsxFilename!,
-						xlsxSheetname!,
-					)}
-			>
-				${t('Save selected items as XLSX')}
-			</button>
-			<slot name="download-menu"></slot>
-		</cosmoz-dropdown-menu>
+		${selectedItems !== All
```
megheaiulian: when()

src/lib/render-footer.ts
```diff
@@ -4,72 +4,112 @@ import { html } from 'lit-html';
...
-		?active=${!isEmpty(selectedItems.length)}
+		?active=${isAllSelected || !isEmpty(selectedCount)}
 		part="bottomBar"
 		exportparts="bar: bottomBar-bar, info: bottomBar-info, buttons: bottomBar-buttons"
 	>
-		<slot name="info" slot="info">
-			${t('{count} selected item', { count: selectedItems.length })}
-		</slot>
+		<span slot="info"> ${allLabel} </span>
+		${selectedItems !== All && enableSelectAll && allSelected
```
megheaiulian: when()

src/lib/render-footer.ts
```diff
@@ -4,72 +4,112 @@ import { html } from 'lit-html';
...
+	saveAsXlsx?: typeof saveAsXlsxAction;
+	setSelectedItems: (
+		items: Item[] | TAll | ((prev: Item[] | TAll) => Item[] | TAll),
+	) => void;
 	csvFilename?: string;
 	xlsxFilename?: string;
 	xlsxSheetname?: string;
 	topPlacement?: string;
+	enableSelectAll?: boolean;
+	allSelected?: boolean;
```
megheaiulian: Why we need allSelected and allItemsCount ?

src/lib/render-footer.ts
```diff
@@ -4,72 +4,112 @@ import { html } from 'lit-html';
...
+				? t('All {count} items selected', { count: allItemsCount })
+				: t('All items selected');
+	} else {
+		allLabel = t('{count} selected item', { count: selectedCount });
+	}
+
+	return html`<cosmoz-bottom-bar
 		id="bottomBar"
-		?active=${!isEmpty(selectedItems.length)}
+		?active=${isAllSelected || !isEmpty(selectedCount)}
```
megheaiulian: This can be determined from  selectedItems ? A local hasSelection variable  computed from selectedItems ?

megheaiulian (changes requested):

src/grouped-list/use-selected-items.ts
basilgood: Removed. We now use flatData directly in deselect, so no extra ref is needed.

src/lib/render-footer.ts
basilgood: Yes, flatData is the flattened render list, so we use it directly instead of re-walking group items.

src/lib/render-footer.ts
basilgood: allSelected is only the header-checked state used to show the footer action; allItemsCount is only for the All {count} items selected label. I kept hasSelection local and removed extra derived state where possible.

src/lib/render-footer.ts
basilgood: I kept hasSelection local and removed extra derived state where possible.

src/lib/render-footer.ts
basilgood: done

.storybook/manager.ts
basilgood: it's not, removed

src/lib/render-footer.ts
```diff
@@ -1,40 +1,52 @@
...
+}: RenderFooterParams) => {
+	const isAllSelected = selectedItems === All;
+	const hasSelection = isAllSelected || selectedItems.length > 0;
+	const showSelectAllItems =
+		selectedItems !== All && enableSelectAll && allSelected;
+	const renderExportMenu = () => {
+		if (selectedItems === All) {
+			return;
+		}
+
```
megheaiulian: We already conditionally call renderExportMenu  if selectedItems is not All /

megheaiulian: Also why did we move it here instead of wrapping in a when  ?

src/lib/use-omnitable.js
```diff
@@ -91,7 +93,11 @@ export const useOmnitable = (host) => {
 		footer: useFooter({
 			host,
 			selectedItems,
+			allSelected: header.allSelected,
+			setSelectedItems,
 			columns,
+			enableSelectAll: host.enableSelectAll === true,
```
megheaiulian: what is this ? enableSelectAll is already boolean ?

src/lib/use-footer.ts
```diff
@@ -1,12 +1,23 @@
...
 	csvFilename?: string;
 	xlsxFilename?: string;
 	xlsxSheetname?: string;
 	topPlacement?: string;
+	allItemsCount?: number;
 }
 
 interface UseFooterParams {
 	host: FooterHost;
+	selectedItems: Item[] | TAll;
```
megheaiulian: Need a TSelecton = Item[] | TAll type otherwise we duplicate it all over the place.

src/lib/render-footer.ts
```diff
@@ -71,5 +83,37 @@ export const renderFooter = ({
...
+	};
+	const allLabel = when(
+		isAllSelected,
+		() =>
+			allItemsCount !== undefined
+				? t('All {count} items selected', { count: allItemsCount })
+				: t('All items selected'),
+		() =>
+			t('{count} selected item', {
+				count: selectedItems === All ? 0 : selectedItems.length,
```
megheaiulian, line 92: selecteditems can't possibly be All in this branch ?

src/lib/render-footer.ts
```diff
@@ -1,40 +1,52 @@
...
-	>
-		<slot name="info" slot="info">
-			${t('{count} selected item', { count: selectedItems.length })}
-		</slot>
-		<slot name="actions" id="actions"></slot>
-		<slot name="bottom-bar-toolbar" slot="bottom-bar-toolbar"></slot>
-		<slot name="bottom-bar-menu" slot="bottom-bar-menu"></slot>
-		<cosmoz-dropdown-menu part="extra" slot="extra" .placement=${topPlacement}>
+	enableSelectAll,
+	allSelected = false,
```
megheaiulian: false is a weird fallback value. let it be undefined.

megheaiulian (changes requested):

src/lib/use-footer.ts
basilgood: done

src/lib/render-footer.ts
basilgood: Yes, you're right!
I kept the condition local and switched it to when(...) to match the rest of the template.
the export branch now guards itself and uses the existing selectedItems directly.

src/lib/use-omnitable.js
basilgood: Yes, it's already boolean.
I removed the explicit === true check and pass host.enableSelectAll through directly.

src/lib/render-footer.ts
basilgood: Yes — the outer when(selectedItems !== All, ...) already guarantees the array case here, so the inner All guard was redundant. I removed it and kept the branch minimal.

src/lib/render-footer.ts
basilgood: I removed this weird fallback. allSelected is now passed as the real header state from use-omnitable.js:
`footer: useFooter({allSelected: header.allSelected})`.

src/grouped-list/use-selected-items.ts
```diff
@@ -60,23 +62,39 @@ export const useSelectedItems = ({
...
+		(item: Item | GroupItem<Item>) => {
+			const groupItem = item as GroupItem<Item>;
+			const items = groupItem.items ?? [item as Item];
+			setSelectedItems((selection) => {
+				if (selection === All) {
+					return (flatData ?? [])
+						.filter((selectedItem) => !isGroup(selectedItem))
+						.filter(
+							(selectedItem) =>
+								!items.some((item) => compareItemsFn(selectedItem, item)),
```
megheaiulian: I think these does not need to be so complicated. we can just check by reference.

megheaiulian (changes requested):
Looks much better. One thing only.

src/grouped-list/use-selected-items.ts
basilgood: I simplified it to a by-reference check: when selection === All, we take flatData that it has group filtered out and now we have only the real items. Nice, thank you for suggestion.

megheaiulian (approved):

PR #1044, opened 2026-07-09
fix: harden filter deserialization for malformed hash values

basilgood:
## Summary
- fix FE-935 by making list filter deserialization resilient to malformed or truncated hash values
- sanitize deserialize error logging to avoid logging raw filter payloads, while preserving useful error metadata
- harden range-based column deserializers (amount/date/datetime/number/time) to safely return null for non-string filters
- harden base column deserialization to return null instead of throwing on malformed percent-escape sequences
- add regression tests for malformed percent-escape sequences and non-string filter deserialization

## Testing
- npm test -- --files test/list.test.js
- npm test -- --files test/range-date.test.js
- npm test -- --files test/basic.test.js

Note: targeted single-file runs can report global coverage-threshold failures even when all tests in the file pass.

### DIFF
--- a/src/cosmoz-omnitable-column-amount.js
+++ b/src/cosmoz-omnitable-column-amount.js
@@ -81,7 +81,7 @@ class OmnitableColumnAmount extends columnMixin(PolymerElement) {
 	}
 
 	deserializeFilter(column, filter) {
-		if (filter == null || filter === '') {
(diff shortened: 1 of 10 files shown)


megheaiulian (approved):
Looks good to me!

cristinecula (approved):

PR #1045, opened 2026-07-10
feat: support inline mode for auto-height omnitable (no internal scrolling)

cristinecula:
## Summary

Adds an `inline` attribute to `cosmoz-omnitable` that enables auto-height mode — the omnitable expands to show all rows without internal scrolling, letting the page scroll instead.

Closes FE-954

## How it works

When `<cosmoz-omnitable inline>` is active:

- `:host([inline])` → `overflow: visible` (remove fixed-height constraint)
- `:host([inline]) .tableContent` → `overflow-y: visible; flex: none` (no internal scrolling)
- `:host([inline]) .tableContent-scroller` → `overflow: visible; flex-basis: auto` (no internal scrolling)

With all parents having `overflow: visible`, lit-virtualizer's default (non-scroller) mode naturally detects no clipping ancestors and falls back to the browser viewport for visibility calculations. The virtualizer sets `minHeight` on `cosmoz-grouped-list` to the total content height, making it expand to show all rows while still virtualizing rendering.

### DIFF
--- a/.changeset/hungry-pets-thank.md
+++ b/.changeset/hungry-pets-thank.md
@@ -1,0 +1,5 @@
+---
+'@neovici/cosmoz-omnitable': minor
+---
+
+Add `inline` attribute for auto-height mode — the omnitable expands to show all rows without internal scrolling, letting the page scroll instead.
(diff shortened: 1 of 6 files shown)


cristinecula:
Can be tested here https://neovici.github.io/cosmoz-omnitable/pr/pr-1045/?path=/story/components-cosmozomnitable--inline

megheaiulian (approved):

PR #1090, opened 2026-08-05
fix: use data-align attribute instead of HTML align on cells

megheaiulian:
## Problem

The HTML `align` attribute is a presentational attribute. Browsers apply `text-align` via the user-agent stylesheet to elements with `align="right"`. Since `cosmoz-omnitable-item-row` uses `useShadowDOM: false`, the cell divs with `align="right"` are in the omnitable's shadow DOM. The user-agent stylesheet's `text-align` was being inherited by the `cosmoz-bottom-bar` (also in the omnitable's shadow DOM), causing the "Selected:" text to shift position by a few subpixels — producing visual regression test failures (647px diffs).

## Fix

Switch from HTML `align` attribute to `data-align` custom attribute. CSS rules updated from `.cell[align='right']` to `.cell[data-align='right']`. This avoids the presentational attribute behavior while maintaining the same CSS specificity.

## Test plan

- [x] `npm test` — 325 passed, 0 failed
- [x] `npm run build` (tsc) — clean
- [x] `npm run lint` — clean
- [ ] Visual tests in cosmoz-frontend pass with this fix

### DIFF
--- a/.changeset/fix-data-align-attribute.md
+++ b/.changeset/fix-data-align-attribute.md
@@ -1,0 +1,5 @@
+---
+'@neovici/cosmoz-omnitable': patch
+---
+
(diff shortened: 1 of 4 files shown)


megheaiulian:
Closing — will discuss approach internally first.

PR #1091, opened 2026-08-05
chore: migrate use-header and use-public-interface to TypeScript

basilgood:
## Summary

Migrate the header logic and public API hooks to TypeScript as part of the ongoing TS migration (PR3 of the sequence tracked in Linear FE-1008). Also replace MIGRATION.md with a trimmed AGENTS.md.

## Files

- `src/lib/use-header.js` → `src/lib/use-header.ts`
- `src/lib/use-public-interface.js` → `src/lib/use-public-interface.ts`
- `MIGRATION.md` removed
- `AGENTS.md` rewritten (codebase conventions + gotchas only)

## Approach

Follows the TypeScript conventions in AGENTS.md:

- `!` non-null assertions and `instanceof`/`Array.isArray`/`typeof` guards over `as` casts
- `querySelector<T>()` type parameter (not a cast) for queried elements
- Local interfaces for host/params shapes; `[key: string]: unknown` index signature for `...rest`/`...api` spread passthrough
- `useProperty<TSelection<Item>>("selectedItems", [])` explicit generic
- Comma-first + tabs preserved
- No runtime behavior changes; memo deps and callback shapes unchanged

## MIGRATION.md → AGENTS.md

MIGRATION.md mixed migration process (PR plan, sequence, workflow) with codebase style. The migration structure is tracked in Linear (FE-1008), not in a committed file — a process doc in the repo goes stale. AGENTS.md now holds only non-obvious knowledge an agent cannot derive from ls/reading the code: code conventions, TypeScript conventions, and gotchas. The Architecture regurgitation, file tables, and Key Files Map were dropped (they go stale — the migration just proved it).

## Verification

- `npm run build` (tsc): clean
- `npm run lint`: 0 errors
- `npm test`: 325 passed, 0 failed

Refs FE-1008


basilgood:
## Note on the codecov/patch check

The `codecov/patch` check is red because two guards added during the migration add dead branches that are not (and cannot be) exercised by the existing tests:

- `use-header.ts` — `if (!(event.target instanceof HTMLInputElement)) return;` in `onAllCheckboxChange`. This handler is bound via `@input` to a checkbox (see `render-header.ts`), so `event.target` is always an `HTMLInputElement`; the `return` never runs.
- `use-public-interface.ts` — `if (!(ev instanceof CustomEvent)) return;` in the `legacy-filter-changed` handler. The event is always dispatched as a `CustomEvent` (see `cosmoz-omnitable-column-mixin.ts`), so the `return` never runs.

These guards are required by the migration rules — "Use `!` and minimal guards over `as` casts" — and they do not change the valid runtime flow (the branches are dead). They were chosen over `as` casts deliberately.

Coverage of the rest of the diff is unchanged from the original `.js`:
- `use-header.ts`: 115/115 lines, 2/2 functions covered (100%).
- `use-public-interface.ts`: the only uncovered functions (`removeItems`, `replaceItem`, `selectAll`, `deselectAll`, `isItemSelected`, `handler`) are the same 6 pre-existing untested public API methods that were already uncovered in the original `use-public-interface.js` (`FNH:6` of `FNF:12` on master).

Overall project coverage: 87.91% (vs 87.96% on master — a 0.05% delta from the new interface/type-only lines). The codecov/patch target of 88.66% is not met solely because of the two dead-branch guards above.

megheaiulian (approved):

